PR #71, opened 2015-01-14
basic support letters functionality

dreamfall:


app/models/support_letter.rb
```diff
@@ -0,0 +1,5 @@
+class SupportLetter < ActiveRecord::Base
+  belongs_to :supporter
+
```
RuslanHamidullin: minor thing: might be worth to add:
```
belongs_to :form_answer
```

RuslanHamidullin: ahh ignore it ^

RuslanHamidullin: I mistaken above ^

RuslanHamidullin:
Looks good to merge :+1:

PR #72, opened 2015-01-14
Removing slimmer

hammerfunk:
Most style is there except the form page nav which doesn't space out correctly.


dreamfall:
Looks good.

PR #73, opened 2015-01-19
Qae1019 collaborators

RuslanHamidullin:


matthewford:
Note, app needs to be functional with js turned off. not sure if fancybox is.

RuslanHamidullin:
That make sence, will correct this one when will have some spare time

PR #77, opened 2015-02-05
Qae1162 collaborators fixes

RuslanHamidullin:
* Switched from using fancybox to simple redirects in views as we need this working with Javascript turned off too
* Added some missing Rspec tests

### DIFF
--- a/app/controllers/account/collaborators_controller.rb
+++ b/app/controllers/account/collaborators_controller.rb
@@ -32,12 +32,22 @@ class Account::CollaboratorsController < Account::BaseController
       account, 
       create_params).run
     self.collaborator = add_collaborator_interactor.collaborator
+
+    if add_collaborator_interactor.success?
+      redirect_to account_collaborators_path, 
+                  notice: "#{collaborator.email} successfuly added to Collaborators!"
+    else
+      render :new
+    end
   end
 
   def destroy
     collaborator.account_id = nil
     collaborator.role = nil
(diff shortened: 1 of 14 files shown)


app/controllers/account/collaborators_controller.rb
```diff
@@ -32,12 +32,22 @@ def create
       account, 
       create_params).run
     self.collaborator = add_collaborator_interactor.collaborator
+
+    if add_collaborator_interactor.success?
+      redirect_to account_collaborators_path, 
+                  notice: "#{collaborator.email} successfuly added to Collaborators!"
```
dreamfall: typo: `successfully`. Same for notice below.

app/views/account/collaborators/new.html.slim
```diff
@@ -0,0 +1,16 @@
+- provide(:page_content_class, "multi-page page-account")
+
+header.page-header.group class="#{'page-header-wider' unless current_user.completed_registration}"
```
dreamfall: Minor thing: no need to use string interpolation
```ruby
class=('page-header-wider' unless current_user.completed_registration)
```

app/controllers/account/collaborators_controller.rb
RuslanHamidullin: ahhha :bomb:

RuslanHamidullin:
Gonna fix, thanks :dancers:

PR #82, opened 2015-02-13
[QAE1020] International-Trade-Award-Form

RuslanHamidullin:
"International Trade Award Form Draft 2" implementation

* Corrects of existing  InternationalTradeForm.

* Some corrects on QueenAwardHolderQuestion:
   1) Fixed UI / js issue with "Remove" button displaying
   2) Added some new attributes in app/views/qae_form/_queen_award_holder_question.html.slim:
       - data-dependable-values
       - data-parent-option-dependable-key
       - data-dependable-option-siffix
   They needs in in terms to realize C1 question requirements.
   In other words:

  If  "A6: QueenAwardHolderQuestion" has items in list for "Internation Trade" (3 or 6 years) - we need   to hide C1 controls and display specific header placeholder and off course change number for inputs for C1 related questions.

   Javascript part of this located in separated js class (app/assets/javascripts/frontend/custom_questions/options_with_preselected_conditions_question.js.coffee)

* Implemented "A13", based on A6 implementation
  By the way added ability to clean up new entry inputs on using "Add Another" js button.
  As by "add-example" template used in related js logic will be prefilled in case if there is at least of one existing entry.
  It's added to application.js file.
  Main idea is in optional "data-need-to-clear-example" data attribute for ".list-add" css selector

* Implemented javascript logic for "B 2.1"
  So it's new type of conditional questions, which should display N number of subquestion entries in list in dependency on selected amount (1,2,3 ... so on)
  For this reason I added optional "data-type" attribute with possible "in_clause_collection" value.

* Corrected displaying of "HeadOfTheBussiness" question block in F section.
  As now we need to use dropdown ("Mr/Mrs so on") with extra "Specify" field

### DIFF
--- a/lib/qae_form_builder/options_with_preselected_conditions_question.rb
+++ b/lib/qae_form_builder/options_with_preselected_conditions_question.rb
@@ -1,0 +1,74 @@
+class QAEFormBuilder
+  class PlaceholderPreselectedCondition
+    attr_accessor :question_key, 
+                  :question_suffix, 
+                  :parent_question_answer_key,
+                  :answer_key,
+                  :question_value, 
+                  :placeholder_text
+
+    def initialize(question_key, question_suffix, parent_question_answer_key, answer_key, question_value, placeholder_text)
+      @question_key = question_key
+      @question_suffix = question_suffix
+      @parent_question_answer_key = parent_question_answer_key
+      @answer_key = answer_key
+      @question_value = question_value
+      @placeholder_text = placeholder_text
+    end
+  end
+
+  class OptionsWithPreselectedConditionsQuestion < OptionsQuestion
+    attr_accessor :main_header, 
+                  :placeholder_preselected_conditions, 
+                  :options, 
+                  :question_key
+
+    def after_create
+      @placeholder_preselected_conditions = []
+      @options = []
+    end
+  end
+
+  class OptionsWithPreselectedConditionsQuestionBuilder < OptionsQuestionBuilder
+    def main_header text
+      @q.main_header = text
+    end
+
+    def placeholder_preselected_condition(q_key, q_suffix, p_answer_key, answer_key, q_value, placeholder_text)
+      @q.question_key = q_key
+      @q.placeholder_preselected_conditions << PlaceholderPreselectedCondition.new(
+        q_key, q_suffix, p_answer_key, answer_key, q_value, placeholder_text
+      )
+    end
+  end
+
+  class OptionsWithPreselectedConditionsQuestionDecorator < QuestionDecorator
+    def linked_answers
+      @linked_answers ||= JSON.parse(answers[delegate_obj.question_key.to_s] || '[]').map do |answer|
+        JSON.parse(answer)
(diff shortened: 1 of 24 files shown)


lib/qae_form_builder/options_with_preselected_conditions_question.rb
```diff
@@ -0,0 +1,74 @@
+class QAEFormBuilder
+  class PlaceholderPreselectedCondition
+    attr_accessor :question_key, 
+                  :question_suffix, 
+                  :parent_question_answer_key,
+                  :answer_key,
+                  :question_value, 
+                  :placeholder_text
+
+    def initialize(question_key, question_suffix, parent_question_answer_key, answer_key, question_value, placeholder_text)
```
dreamfall: To many arguments. Better to pass a hash.

RuslanHamidullin: Yeah, right. Gonna correct

lib/qae_form_builder/options_with_preselected_conditions_question.rb
```diff
@@ -0,0 +1,74 @@
...
+      end
+    end
+
+    def preselected_condition
+      @preselected_condition ||= placeholder_preselected_conditions.select do |c|
+        linked_answers.select do |a|
+          a[c.question_suffix.to_s] == c.parent_question_answer_key
+        end.present?
+      end.first
+    end
```
dreamfall:
```ruby
    def preselected_condition
      @preselected_condition ||= placeholder_preselected_conditions.detect do |c|
        linked_answers.any? do |a|
          a[c.question_suffix.to_s] == c.parent_question_answer_key
        end
      end
    end
```

RuslanHamidullin: right, that make sence

lib/qae_form_builder/options_with_preselected_conditions_question.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    def placeholder_preselected_conditions
+      @placeholder_preselected_conditions ||= delegate_obj.placeholder_preselected_conditions
+    end
+
+    def preselected_condition_by_option(option)
+      placeholder_preselected_conditions.select do |condition|
```
dreamfall: Same here, better use `detect`

lib/qae_form_builder/subsidiaries_associates_plants_question.rb
```diff
@@ -6,4 +6,11 @@ class SubsidiariesAssociatesPlantsQuestionBuilder < QuestionBuilder
   class SubsidiariesAssociatesPlantsQuestion < Question
   end
 
+  class SubsidiariesAssociatesPlantsQuestionDecorator < MultiQuestionDecorator
```
dreamfall: no need to inherit this class from `MultiQuestionDecorator`

It has only `awards` method (i renamed it to `entities` in my branch so it can be universal)

PR #83, opened 2015-02-13
Ep form draft part two

dreamfall:


app/assets/javascripts/application.js.coffee
```diff
@@ -415,13 +416,17 @@ jQuery ->
...
       if can_add
-        add_example = add_example.replace(/(form\[\w+\]\[)(\d+)\]/g, "$1#{list_size+1}]")
+        add_example = add_example.replace(/(form\[(\w+|_)\]\[)(\d+)\]/g, "$1#{list_size+1}]")
         question.find(".list-add").append("<li>#{add_example}</li>")
 
         need_to_clear_example = question.find(".list-add").attr("data-need-to-clear-example")
         if (typeof(need_to_clear_example) != typeof(undefined) && need_to_clear_example != false)
           clearFormElements(question.find(".list-add li:last"))
 
+        # charcount needs to be reinitialized
```
RuslanHamidullin: small issue I found in chrome console in case if selector not found.
maybe w
error
error2

e can check it
```
char_counter_div = question.find(".list-add > li:last .js-char-count")
if char_counter_div.length > 1
  question.find(".list-add > li:last .js-char-count").removeCharcountElements()
  question.find(".list-add > li:last .js-char-count").charcount()
```

app/views/qae_form/_award_holder_question.html.slim
```diff
@@ -1,28 +1,11 @@
-ul.list-add data-add-limit="10"
-  input name="form[awards][array]" value="true" type="hidden"
+ul.list-add data-need-to-clear-example=true
```
RuslanHamidullin: Not sure but suppose you need to add data-add-limit="<Max number>" here too
in terms to limit awards count and apply "disabling add another button" js  when max is reached

RuslanHamidullin: Otherwise, unlimited awards can be added

RuslanHamidullin:
Looks nice :+1:

PR #85, opened 2015-02-16
Admin login/signup pages restyled/restructured

hammerfunk:
Restructured the admin login/signup pages to be more inline with the user's login/signup page designs


RuslanHamidullin:
Looks good for me :+1:

PR #92, opened 2015-02-18
Add form answer state machine.

jaleszek:
Initial implementation of the application state machine.
All transitions are permitted for now as flow is not sorted yet.
Transition triggers need to be added in appropriate places.
States are grouped with indices which represents the phase (based on documentation).

My thinking is to use it for filtering purposes first and then add the transitions triggers to relevant places.
Transitions table stores all historical data.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -37,6 +37,7 @@ gem 'sinatra', require: false
 gem 'foreman'
 gem 'jquery.fileupload-rails'
 gem 'carrierwave'
+gem 'statesman'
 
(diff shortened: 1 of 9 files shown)


RuslanHamidullin:
Looks clear for me :+1:

RuslanHamidullin:
@dreamfall any notes?

dreamfall:
@RuslanHamidullin I'm not 100% sure about the number of states. Seems kind of complex.

PR #94, opened 2015-02-18
Use passenger standalone in development

matthewford:
So we are closer to production.


RuslanHamidullin:
Look reasonable for me :+1:

PR #95, opened 2015-02-18
Update Gems to be Rails 4 compatible

matthewford:
Also need to move jobs to active job format (including delayed mailer) to make background jobs agnostic so we can change the backend.


config/initializers/assets.rb
```diff
@@ -3,6 +3,9 @@
 # Version of your assets, change this if you want to expire all your assets.
 Rails.application.config.assets.version = '1.0'
 
+# Add additional assets to the asset load path
+# Rails.application.config.assets.paths << Emoji.images_path
+
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
 Rails.application.config.assets.precompile += [ "application-admin.css", "application-admin.js" ]
```
RuslanHamidullin: I suppose it would be:
```
Rails.application.config.assets.precompile += %w{
     application.scss
     application.js.coffee
     application-admin.scss
     application-admin.js.coffee
     application-ie6.scss
     application-ie7.scss
     application-ie8.scss
     application-ie9.scss
 ]
```

RuslanHamidullin: And then remove similar declaration from config/application.rb

RuslanHamidullin:
Good to merge :+1:  Gonna then correct config/initizalizers/assets.rb

PR #97, opened 2015-02-19
Restyle admin users and table filters

hammerfunk:


dreamfall:
Looks good, but can't be merged.

PR #100, opened 2015-02-19
Placeholder page for admin dashboard

hammerfunk:


RuslanHamidullin:
Good to merge :+1:

app/controllers/content_only_controller.rb
```diff
@@ -9,4 +9,7 @@ def dashboard
     load_eligibilities
     @user_award_forms = current_user.account.form_answers.order("award_type")
   end
+
+  def admin_dashboard
```
RuslanHamidullin: minor thing: looks like this one method is not used 
```
def admin_dashboard
end
```
we can remove it

PR #101, opened 2015-02-19
Applications list

jaleszek:


Gemfile
```diff
@@ -96,6 +96,8 @@ group :development do
   gem 'quiet_assets'
   gem 'rack-mini-profiler', require: false
   gem 'passenger'
+  gem 'better_errors'
```
matthewford: Is this better than the Rails 4.2 web console (which also uses binding_of_caller under the covers)?

matthewford: Added here https://github.com/bitzesty/qae/blob/applications-list/Gemfile#L106 (I don't think we need both)

jaleszek: I like better errors because it sets up the debugger directly in the place where exception happened what is crucial with fast debugging. Moreover it has easy context switching to debug full execution path.

matthewford: Lets remove the other gem then

PR #124, opened 2015-02-25
Csv reports

jaleszek:
CSV reports functionality. It's initial sketch of that as a lot of the requirements wait for clarification.


app/controllers/admin/reports_controller.rb
```diff
@@ -0,0 +1,16 @@
+class Admin::ReportsController < Admin::BaseController
+  def show
+    authorize :report, :show?
+
+    respond_to do |format|
+      format.html
+      format.csv{ send_data resource.build }
```
matthewford: Space missing to the left of {.

app/models/form_answer.rb
```diff
@@ -107,6 +107,18 @@ def promotion?
     award_type == 'promotion'
   end
 
+  def trade?
+    award_type == 'trade'
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/form_answer.rb
```diff
@@ -107,6 +107,18 @@ def promotion?
...
+  end
+
+  def development?
+    award_type == 'development'
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/form_answer.rb
```diff
@@ -107,6 +107,18 @@ def promotion?
...
+  end
+
+  def innovation?
+    award_type == 'innovation'
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

app/models/reports/admin_report_builder.rb
```diff
@@ -0,0 +1,143 @@
...
+      end
+    end
+
+    csv_string
+  end
+
+  private
+
+  def headers
+    MAPPING.map{|m| m[:label]}
```
matthewford: Space missing to the left of {.<br>Space between { and | missing.<br>Space missing inside }.

app/models/form_answer.rb
```diff
@@ -104,7 +104,19 @@ def document
   end
 
   def promotion?
-    award_type == 'promotion'
+    award_type == "promotion"
```
RuslanHamidullin: I think here would be great to use favorite Vasily's 'enumerize' (https://github.com/brainspec/enumerize) gem.
Would be:
```
enumerize :award_type, in: POSSIBLE_AWARDS, predicates: true
```
instead of:
```
def promotion?
    award_type == "promotion"
  end

  def trade?
    award_type == "trade"
  end

  def development?
    award_type == "development"
  end

  def innovation?
    award_type == "innovation"
  end
````

**Comment by RuslanHamidullin** [comment]
Looks nice :+1: Just small note about possible using of "enumerize" gem for predicates

File: app/models/form_answer.rb (hunk 6 above)
Comment (matthewford) [inline app/models/form_answer.rb]: if its simple enum, its in rails 4 now. http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

Comment (RuslanHamidullin) [inline app/models/form_answer.rb]: looks like it, @dreamfall would do you think about Rails's enum http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html
instead of enumerize ?

Comment (matthewford) [inline app/models/form_answer.rb]: would need to change the type to an int

Comment (dreamfall) [inline app/models/form_answer.rb]: Well, ActiveRecord's Enum does not provide a whole bunch of features that Enumerize does (using strings as enum values /requires more space oh HDD for DB, but helps to avoid migrations when enumed field was a string like in our case/, I18n support, built in scopes and possibility of customizing Value class that will be instantiated and returned when for example `FormAnswer#award_type` is called). 

When all you need is methods like `FormAnswer#promotion?` then `ActiveRecord::Enum` seems to be a good choice. In most of the other cases enumerize is both more powerful and flexible.

Comment (RuslanHamidullin) [inline app/models/form_answer.rb]: enumerize +1

File: app/models/form_answer.rb [hunk 7]
```diff
@@ -123,6 +135,10 @@ def company_or_nominee_from_document
     name.presence
   end
 
+  def fill_progress_in_percents
+    "#{((fill_progress || 0) * 100).round.to_i}%"
```
dreamfall: Small note: don't think`to_i` is needed here.

app/policies/report_policy.rb
```diff
@@ -0,0 +1,5 @@
+class ReportPolicy < ApplicationPolicy
+  def show?
+    true
```
dreamfall: I'm not sure if (lead) assessors should be able to generate this reports.

matthewford: confirm with Michael but I think this is admins only

app/models/form_answer.rb
jaleszek: Tbh I wouldn't like to change the type of the `award_type` column as it's coupled with existing code in few places `63 matches across 22 files` so probably will just add enumerize gem. Thanks

RuslanHamidullin: :+1:

app/models/form_answer.rb
jaleszek: sure, just extracted this method from another place

dreamfall: It would be great if you clean up the code you copied it from as well if you remember where it is. :heart:

jaleszek: I would love to live in ideal world : )

PR #127, opened 2015-02-25
Upload Document and Links fixes

RuslanHamidullin:
1) doesn't work "Remove" of uploaded entry if just uploaded and didn't refresh page (A15 Innovation Form)
2) Multiple Uploads - when try to upload 2 files and then save -> system saves just one of them (E1 Innovation Form)


dreamfall:
All good!

PR #128, opened 2015-02-25
Updated the admin user edit page to show all fields like frontend

hammerfunk:
Also we possibly need it to work with auto-saving. I don't know how we'll make adding a collaborator work without it since it opens a new page. OR each tab needs to open it's own page.
This needs discussion about UX.

The collaborators section needs work which you can find with a TODO:
- Should use a better way to reference the current users `collaborators`
- Adding a new collaborator
- Removing a collaborator

### DIFF
--- a/app/assets/stylesheets/application-admin.scss
+++ b/app/assets/stylesheets/application-admin.scss
@@ -10,7 +10,7 @@
 $error-colour: #df3034;
 $yellow: #ffbf47;
 $button-colour: #00823b;
-@import 'frontend/password_strength';
+@import "frontend/password_strength";
 
 @import "admin/_toggle";
 @import "admin/layout";
--- a/app/assets/stylesheets/application-admin.scss
+++ b/app/assets/stylesheets/application-admin.scss
@@ -2,8 +2,16 @@
  *= require_self
  */
 
+// Bootstrap
 @import "bootstrap-sprockets";
 @import "bootstrap";
+
+// Password strength bar
+$error-colour: #df3034;
+$yellow: #ffbf47;
+$button-colour: #00823b;
+@import 'frontend/password_strength';
+
 @import "admin/_toggle";
 @import "admin/layout";
 @import "admin/base";
--- a/app/assets/stylesheets/frontend/password_strength.scss
+++ b/app/assets/stylesheets/frontend/password_strength.scss
@@ -33,9 +33,9 @@ $colorLevel4: $button-colour;
     transition: width 0.2s ease-in-out;
   }
   .password-meter-text {
+    position: absolute;
     top: 50%;
     right: 0;
-    position: absolute;
     width: 70px;
     margin-top: -0.5em;
     color: #333333;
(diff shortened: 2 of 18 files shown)


app/assets/stylesheets/application-admin.scss
```diff
@@ -2,8 +2,16 @@
...
 
+// Bootstrap
 @import "bootstrap-sprockets";
 @import "bootstrap";
+
+// Password strength bar
+$error-colour: #df3034;
+$yellow: #ffbf47;
+$button-colour: #00823b;
+@import 'frontend/password_strength';
```
matthewford: Prefer double-quoted strings

app/assets/stylesheets/frontend/password_strength.scss
```diff
@@ -34,12 +33,13 @@ $colorLevel4: $button-colour;
     transition: width 0.2s ease-in-out;
   }
   .password-meter-text {
+    top: 50%;
```
matthewford: Properties should be ordered position, top, right, width, margin-top, color, font-size

app/views/account/collaborators/_item.html.slim
```diff
@@ -1,7 +1,8 @@
-li id="#{dom_id(collaborator)}"
-  = link_to "Remove", account_collaborator_path(collaborator),
-                      method: :delete,
-                      class: 'remove-link'
+li.list-group-item id="#{dom_id(collaborator)}"
```
dreamfall: `id=dom_id(collaborator)` works as well

hammerfunk: Good spot!

app/views/admin/users/_fields_user_details.html.slim
```diff
@@ -0,0 +1,53 @@
...
+                label_html: { class: "visuallyhidden" }
+
+fieldset.question-group
+  h3 Email
+  .row
+    .col-sm-4
+      = f.input :email,
+                wrapper_html: { class: "form-group" },
+                input_html: { class: "form-control" },
+                label_html: { class: "visuallyhidden" }
```
dreamfall: don't think it's worth changing but fyi:
You can pass `label: nil` in options to remove a label.

app/views/admin/users/_fields_collaborators.html.slim
```diff
@@ -0,0 +1,5 @@
+/ TODO collaborators new/delete, also the collaborators variable
+- if params[:action] == "edit"
```
dreamfall: From RoR guides: The params hash will always contain the :controller and :action keys, but you should use the methods controller_name and action_name instead to access these values. 

Same goes for other usages of `params[:action]` in your code.

app/views/admin/users/_fields_user_details.html.slim
hammerfunk: We don't want to remove it (screenreaders), just visually. Otherwise yeah, I would've done `label: false`

PR #131, opened 2015-02-26
Add searching for prefixes in application list

jaleszek:
To use prefixes (not only entire words) for searching.

### DIFF
--- a/app/models/form_answer.rb
+++ b/app/models/form_answer.rb
@@ -4,14 +4,21 @@ class FormAnswer < ActiveRecord::Base
   include PgSearch
   extend Enumerize
 
-  pg_search_scope :basic_search, against: [
-    :urn,
-    :award_type_full_name,
-    :award_year,
-    :company_or_nominee_name,
-    :nominee_full_name,
-    :user_full_name
-  ]
+  pg_search_scope :basic_search,
+                  against: [
+                    :urn,
+                    :award_type_full_name,
+                    :award_year,
+                    :company_or_nominee_name,
+                    :nominee_full_name,
+                    :user_full_name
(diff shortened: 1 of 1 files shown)


app/models/form_answer.rb
```diff
@@ -4,14 +4,21 @@ class FormAnswer < ActiveRecord::Base
...
-  pg_search_scope :basic_search, against: [
-    :urn,
-    :award_type_full_name,
-    :award_year,
-    :company_or_nominee_name,
-    :nominee_full_name,
-    :user_full_name
-  ]
+  pg_search_scope :basic_search,
+    against: [
```
matthewford: Align the parameters of a method call if they span more than one line.

app/models/form_answer.rb
```diff
@@ -4,14 +4,22 @@ class FormAnswer < ActiveRecord::Base
...
-  pg_search_scope :basic_search, against: [
-    :urn,
-    :award_type_full_name,
-    :award_year,
-    :company_or_nominee_name,
-    :nominee_full_name,
-    :user_full_name
-  ]
+  pg_search_scope
+    :basic_search,
```
matthewford: unexpected token tCOMMA

app/models/form_answer.rb
```diff
@@ -4,14 +4,22 @@ class FormAnswer < ActiveRecord::Base
...
+    against: [
+      :urn,
+      :award_type_full_name,
+      :award_year,
+      :company_or_nominee_name,
+      :nominee_full_name,
+      :user_full_name
+    ],
```
matthewford: unexpected token tCOMMA

PR #135, opened 2015-02-26
Change CSV export to new requirements

jaleszek:
New bunch of methods for CSV report.

### DIFF
--- a/lib/qae_form_builder/question.rb
+++ b/lib/qae_form_builder/question.rb
@@ -12,7 +12,7 @@ class QAEFormBuilder
 
     def input_value options = {}
       result = if options[:index]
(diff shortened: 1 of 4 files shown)


lib/qae_form_builder/question.rb
```diff
@@ -12,7 +12,7 @@ def input_name options = {}
 
     def input_value options = {}
       result = if options[:index]
-        suffix = options.fetch(:suffix)  
+        suffix = options.fetch(:suffix)
```
matthewford: Use 2 (not -7) spaces for indentation.

jaleszek: I think we can ignore that

PR #138, opened 2015-02-27
[By The Way Fixes]

RuslanHamidullin:
* Fixed frontend validation for financial years
It didn't work and also caused sometimes blocking of the form
without any message


hammerfunk:
Beautiful :+1:

PR #143, opened 2015-03-02
Assign assessor to category

jaleszek:
Need to use many to many association to store the role as it's per category.
Categories are the award types, for now just used the constant from FormAnswer but possibly can be normalized in the future.


app/models/assessor.rb
```diff
@@ -7,10 +7,9 @@ class Assessor < ActiveRecord::Base
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :trackable, :validatable, :confirmable
 
-  validates :first_name, :last_name, :role, presence: true
+  validates :first_name, :last_name,  presence: true
```
matthewford: Unnecessary spacing detected.

PR #144, opened 2015-03-02
Assign assessor to category

jaleszek:
Need to use many to many association to store the role as it's per category.
Categories are the award types, for now just used the constant from FormAnswer but possibly can be normalized in the future.


RuslanHamidullin:
All clear :+1:

PR #190, opened 2015-03-17
Secureheaders

matthewford:


RuslanHamidullin:
Good :+1:  Looks like need to rebase with master

PR #202, opened 2015-03-19
Lead assessors on admin application sidebar

hammerfunk:


app/decorators/form_answer_decorator.rb
```diff
@@ -117,4 +117,14 @@ def financial_summary_updated_by
   def financial_summary_updated_at
     object.financial_data && object.financial_data["updated_at"]
   end
+
+  def lead_assessors
+    award_leads = Assessor.all.where("#{object.decorate.award_type_slug}_role" => "lead")
+
+    if award_leads.any?
+      award_leads.map { |lead| lead.full_name }.join(", ")
```
matthewford: Pass `&:full_name` as an argument to `map` instead of a block.

app/decorators/form_answer_decorator.rb
```diff
@@ -117,4 +117,14 @@ def financial_summary_updated_by
   def financial_summary_updated_at
     object.financial_data && object.financial_data["updated_at"]
   end
+
+  def lead_assessors
+    award_leads = Assessor.all.where("#{object.decorate.award_type_slug}_role" => "lead")
```
jaleszek: I would write something like: (didn't test)
award_leads = Assessor.leads_for(object.award_type)

class Assessor < ..
...
def self.leads_for(category)
  where(role_meth(category) => "lead")
end

hammerfunk: Tested works great! Also this code looks much cleaner! Thanks

PR #205, opened 2015-03-19
[Security Headers] let's disable it on local

RuslanHamidullin:
I suppose that we can disable gem 'secure_headers'.
We do not need it on dev env.
for Dev env as it adds a lot of warning in chrome /firebug consoles and also increases time on loading pages for dev.
I also thinking that it makes angry frontend guys when they try to debug smth in chrome console.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,9 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  ensure_security_headers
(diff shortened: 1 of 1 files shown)


app/controllers/application_controller.rb
```diff
@@ -5,7 +5,9 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  ensure_security_headers
+  unless (Rails.env.test? || Rails.env.development?)
```
matthewford: Don't use parentheses around the condition of an `unless`.

PR #223, opened 2015-03-26
Eligibility transitions

jaleszek:
Changes state of eligibility every eligibility form update.
Allowed transitions for User are:
`not_eligible -> application_in_progress`
`application_in_progress -> not_eligible`
only.


jaleszek:
@dreamfall would you like to check ?

dreamfall:
looks good

PR #224, opened 2015-03-26
Don't remove the withdrawn apps from scope.

jaleszek:
Apps with state withdrawn should be considered as normal apps, without excluding anything. Actually it breaks app resource when withdrawn state set up.


jaleszek:
good to merge ? anyone ?

dreamfall:
@jaleszek why do we need it?

jaleszek:
Apps with state withdrawn should be considered as normal apps, without excluding anything. Actually it breaks app resource when withdrawn state set up.

PR #227, opened 2015-03-27
press summary for promotion awards / token based authentication 

dreamfall:
add token

### DIFF
--- a/app/controllers/users/press_summaries_controller.rb
+++ b/app/controllers/users/press_summaries_controller.rb
@@ -11,7 +11,7 @@ class Users::PressSummariesController < Users::BaseController
   def update
     @press_summary.reviewed_by_user = true
 
-    if @press_summary.update_attributes(press_summary_params)
+    if @press_summary.update(press_summary_params)
       flash.notice = "Press Comment successfully updated"
       redirect_to action: :show, token: params[:token]
     else
@@ -21,10 +21,10 @@ class Users::PressSummariesController < Users::BaseController
 
   def update_acceptance
     if params[:form_answer][:accepted] == "true"
-      form_answer.update_column(:accepted, true)
+      form_answer.update(accepted: true)
       redirect_to action: :show, token: params[:token]
     else
-      form_answer.update_column(:accepted, false)
+      form_answer.update(accepted: false)
       redirect_to root_url
     end
   end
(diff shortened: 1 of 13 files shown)


spec/mailers/users/winners_press_release_spec.rb
```diff
@@ -33,7 +34,7 @@
     it "renders the body" do
       expect(mail.body.encoded).to match(user.decorate.full_name)
       expect(mail.body.encoded).to have_link("Review Press Comment",
-                                             href: users_form_answer_press_summary_url(form_answer))
+                                             href: users_form_answer_press_summary_url(form_answer, token: press_release.token))
```
matthewford: Line is too long. [128/100]

app/models/press_summary.rb
```diff
@@ -2,11 +2,17 @@ class PressSummary < ActiveRecord::Base
...
   validates :form_answer, :body, presence: true
+  validates :first_name, :last_name, :email, :phone_number, presence: true, if: :reviewed_by_user?
 
+  before_create :set_token
   after_save :notify_applicant
 
   private
 
+  def set_token
+    self.token = SecureRandom.base64(24)
```
jaleszek: do we need any validations constraints for token ?

app/controllers/users/press_summaries_controller.rb
```diff
@@ -1,31 +1,34 @@
 class Users::PressSummariesController < Users::BaseController
```
jaleszek: Maybe just inherit it from ApplicationsController ?

jaleszek: nope, ignore that

app/controllers/users/press_summaries_controller.rb
```diff
@@ -1,31 +1,34 @@
...
-      redirect_to action: :show
+      redirect_to action: :show, token: params[:token]
     else
       render :show
     end
   end
 
+  def update_acceptance
+    if params[:form_answer][:accepted] == "true"
+      form_answer.update_column(:accepted, true)
```
jaleszek: Skipping validations is necessary here ?

dreamfall: @jaleszek no. But there's no point in validating form answer after the flag is changed.

jaleszek:
:+1:  looks good, just let me know about your thoughts about validating token

PR #250, opened 2015-04-01
Fix issue with form multiedit

jaleszek:
+      # Because every text field has separated submit form button
+      # but there is only single huge form for all of the descriptions
+      # it's needed to updated only description marked explicitly by the admin
+      # to achieve data other description fields should be removed from params

https://podio.com/bitzesty/app-qae/apps/stories/items/175 (last comment with pics)

### DIFF
--- a/spec/support/shared_contexts/appraisal_form_context.rb
+++ b/spec/support/shared_contexts/appraisal_form_context.rb
@@ -28,6 +28,14 @@ shared_context "successful appraisal form edition" do
       assert_description_change(secondary, secondary_header)
       assert_description_change(moderated, moderated_header)
     end
+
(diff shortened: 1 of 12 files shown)


spec/support/shared_contexts/appraisal_form_context.rb
```diff
@@ -111,6 +119,31 @@ def assert_description_change(section_id, header_id)
   visit show_path
 end
 
+def assert_multiple_description_change(section_id, header_id)
```
matthewford: Assignment Branch Condition size for assert_multiple_description_change is too high. [34.06/15]

RuslanHamidullin:
Good :+1:

PR #252, opened 2015-04-01
Clarify the deadlines stuff

jaleszek:
- Deadlines clarification, - triggering the automatic states transitions, displaying the hidden sections of admin panel after the deadline.
- Removed the validation from the form_answer, as there can be a 2 sources of the eligibility changes: eligibility form and changes made by the admin (only admin).

### DIFF
--- a/app/models/form_answer_state_machine.rb
+++ b/app/models/form_answer_state_machine.rb
@@ -34,17 +34,12 @@ class FormAnswerStateMachine
   end
 
   def self.trigger_deadlines
-    ends = Deadline.with_states_to_trigger.includes(:settings)
     relevant_states = [
       "submitted",
       "application_in_progress"
     ]
-
-    ends.each do |deadline|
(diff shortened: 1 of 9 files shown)


app/models/form_answer_state_machine.rb
```diff
@@ -127,13 +122,7 @@ def trigger_eligibility_change
...
 
   def permitted_states_with_deadline_constraint
-    year = object.award_year
-    settings = Settings.for_year(year - 1)
-    deadline = settings.deadlines.where(kind: "submission_end").first
-
-    # time restrictions
-    return [] if deadline.trigger_at.blank?
-    if DateTime.now < deadline.trigger_at
+    unless Settings.after_current_submission_deadline?
```
matthewford: Do not use `unless` with `else`. Rewrite these with the positive case first.

RuslanHamidullin:
Looks fine :+1:

PR #270, opened 2015-04-07
Admin docs section

RuslanHamidullin:
Docs section corrects:
1) APPLICATION & SUPPORTING DOCS
```
should have any supporting docs attached to application form (i.e. organisation chart etc. etc.)
```
2) POST SHORTLISTING DOCS
```
This should have audit cert uploaded by user and place to download case summary
```
TODO: Pdf generation for Case Summary

3) OTHER_DOCS
```
"Other Docs" should have anything uploaded by an assessor and admin
```

### DIFF
--- a/app/views/admin/form_answers/docs/_application_and_supporting_docs.html.slim
+++ b/app/views/admin/form_answers/docs/_application_and_supporting_docs.html.slim
@@ -1,0 +1,10 @@
+.document-list
+  p.p-empty class="#{'visuallyhidden' if @form_answer.form_answer_attachments.any?}"
+    ' No documents have been attached to this case.
+  ul.list-unstyled.list-actions
+    - if @form_answer.form_answer_attachments.uploaded_by_user.any?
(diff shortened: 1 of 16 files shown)


app/views/admin/form_answers/docs/_application_and_supporting_docs.html.slim
```diff
@@ -0,0 +1,10 @@
+.document-list
+  p.p-empty class="#{'visuallyhidden' if @form_answer.form_answer_attachments.any?}"
```
jaleszek: you can always write: `p.p-empty class=("visuallyhidden" if @form_answer.form_answer_attachments.any?)`
#justsaying #dontChangeThis

RuslanHamidullin: that make sence, I think I need to correct it as condition also should check support letters

RuslanHamidullin: gonna fix

PR #272, opened 2015-04-07
Automatic tests

jaleszek:


spec/factories/form_answer_factory.rb
```diff
@@ -26,5 +26,16 @@
     trait :submitted do
       submitted true
     end
+
+    trait :with_audit_certificate do
+      document do
+        file = "#{Rails.root}/spec/fixtures/form_answer_document_with_review_audit_certificate_criteria.json"
```
matthewford: Line is too long. [109/100]

jaleszek: No, is not !

PR #301, opened 2015-04-10
Updated the style of error pages

hammerfunk:


hammerfunk:
Yeah...?

Doing this task: Make custom error pages if we're offline with QAE Contact details

But just read again and saw priority was low...

matthewford:
You're replacing key copy in the error message type - they are not all 400's

hammerfunk:
I've left the content but added stuff around it

public/422.html
```diff
@@ -2,14 +2,13 @@
 <html>
   <head>
     <meta http-equiv="content-type" content="text/html; charset=UTF-8">
-    <title>The change you wanted was rejected (422)</title>
```
matthewford: @hammerfunk check again. This line should not have been changed.

hammerfunk:
Good spot! Thanks

PR #303, opened 2015-04-10
Review audit certificate mailing

jaleszek:


app/mailers/assessors/general_mailer.rb
```diff
@@ -5,4 +5,13 @@ def led_application_withdrawn(form_answer_id, assessor_id)
     @subject = "TODO: copy"
     mail to: @assessor.email, subject: @subject
   end
+
+  def audit_certificate_uploaded(form_answer_id)
+    @form_answer = FormAnswer.find(form_answer_id)
+    @assessor = @form_answer.assessors.primary
+    if @assessor.present?
```
RuslanHamidullin: @jaleszek I suppose that need to check 
```
if @assessor.present?
   mail ...
end
```
On level where you call:
```
Assessors::GeneralMailer.audit_certificate_uploaded(form_answer.id).deliver_later!
```
I assume that can be expection in case if  @assessor missing

PR #309, opened 2015-04-10
whitespace

matthewford:


app/pdf_generators/qae_pdf_forms/general/step_pointer.rb
```diff
@@ -33,9 +33,9 @@ def render!
   end
 
   def render_header
-    form_pdf.text step.complex_title, style: :bold, 
-                                  size: 18, 
+    form_pdf.text step.complex_title, style: :bold,
+                                  size: 18,
```
matthewford: Align the elements of a hash literal if they span more than one line.

PR #340, opened 2015-04-14
Fix URN generation.

jaleszek:
Makes the URN sequence global.
Criteria:
QA0001/14T = 1st trade app
QA0002/14S = 1st sustainable development app
QA0003/14T = 2nd trade app

### DIFF
--- a/app/models/form_answer.rb
+++ b/app/models/form_answer.rb
@@ -188,7 +188,7 @@ class FormAnswer < ActiveRecord::Base
     return unless submitted
     return unless award_type
 
-    next_seq = self.class.connection.select_value("SELECT nextval(#{ActiveRecord::Base.sanitize("urn_seq_#{award_type}")})")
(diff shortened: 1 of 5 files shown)


app/models/form_answer.rb
```diff
@@ -188,7 +188,7 @@ def set_urn
     return unless submitted
     return unless award_type
 
-    next_seq = self.class.connection.select_value("SELECT nextval(#{ActiveRecord::Base.sanitize("urn_seq_#{award_type}")})")
+    next_seq = self.class.connection.select_value("SELECT nextval(#{ActiveRecord::Base.sanitize("urn_seq")})")
```
matthewford: Line is too long. [110/100]<br>Prefer single-quoted strings inside interpolations.

RuslanHamidullin:
Looks fine for me, need to do not forget to run rebuild urn task on all 4 servers

PR #352, opened 2015-04-15
CSV reports

jaleszek:
Implementation of structures for all reports.
Implementation of data picking for `registered users` CSV export.

### DIFF
--- a/app/models/reports/form_answer.rb
+++ b/app/models/reports/form_answer.rb
@@ -90,6 +90,22 @@ class Reports::FormAnswer
     end
   end
 
+  def business_region
+    if business_form?
+      doc "principal_address_region"
+    else
+      doc "nominee_personal_address_region"
+    end
+  end
+
+  def employees
+    unless trade?
+      doc("employees_5of5").presence || doc("employees_2of2")
+    else
+      doc("employees_6of6").presence || doc("employees_3of3")
+    end
+  end
+
   def qao_permission
     obj.user.subscribed_to_emails
   end
@@ -129,6 +145,10 @@ class Reports::FormAnswer
 
   def business_form?
     obj.trade? || obj.innovation? || obj.development?
+  end
+
+  def trade?
+    obj.trade?
   end
 
   def promotion?
--- a/spec/models/reports/registered_users_spec.rb
+++ b/spec/models/reports/registered_users_spec.rb
@@ -1,9 +1,9 @@
 require "rails_helper"
 
-describe Reports::AdminReportBuilder do
+describe Reports::RegisteredUsers do
   let!(:form_answer) { create(:form_answer) }
 
-  let(:output) { described_class.new.build }
+  let(:output) { described_class.new(Settings.current).build }
   it "creates the output with relevant data" do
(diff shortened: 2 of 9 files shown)


app/models/reports/form_answer.rb
```diff
@@ -90,6 +90,18 @@ def business_sector_other
...
+  def business_region
+    if business_form?
+      doc "principal_address_region"
+    else
+      doc "nominee_personal_address_region"
+    end
+  end
+
+  def employees
+    doc("employees_5of5").presence || doc("employees_2of2")
```
RuslanHamidullin: @jaleszek for International Trade can be 6 or 3, so need to correct I guess
https://github.com/bitzesty/qae/blob/master/app/forms/qae_2014_forms/international_trade/international_trade_step3.rb#L73

RuslanHamidullin:
@jaleszek other stuff looks good, but need to check Trade Form note above ^

app/models/reports/form_answer.rb
```diff
@@ -90,6 +90,22 @@ def business_sector_other
...
+  def business_region
+    if business_form?
+      doc "principal_address_region"
+    else
+      doc "nominee_personal_address_region"
+    end
+  end
+
+  def employees
+    unless trade?
```
matthewford: Do not use `unless` with `else`. Rewrite these with the positive case first.

jaleszek:
@RuslanHamidullin nice, thanks, fixed.

RuslanHamidullin:
@jaleszek cool, waiting for travis

app/models/reports/form_answer.rb
```diff
@@ -90,6 +90,22 @@ def business_sector_other
...
+      doc("employees_5of5").presence || doc("employees_2of2")
```
dreamfall: Not sure if this is a right check. If user filled stat for 5/6 years first, and then changed it to 2 years `employees_5of5` won't be cleared.

RuslanHamidullin: @jaleszek looks like @dreamfall cacthed another issue.
Yeah, if user changed from 2 to 5 so on, will be both in document

spec/models/reports/registered_users_spec.rb
```diff
@@ -1,9 +1,9 @@
 require "rails_helper"
 
-describe Reports::AdminReportBuilder do
+describe Reports::RegisteredUsers do
   let!(:form_answer) { create(:form_answer) }
 
-  let(:output) { described_class.new.build }
+  let(:output) { described_class.new(Settings.current).build }
```
dreamfall: You're passing `Settings` object here, but code uses `AwardYear` object.

app/models/reports/form_answer.rb
RuslanHamidullin: So need to check selected value for each type of form.
1) Trade 
 "trade_commercial_success" :
```
3 to 5
6 plus
```
2) Sustainable dev form:
"development_performance_years"
```
2 to 4
5 plus
```
3) Innovation
"innovation_performance_years"
```
2 to 4
5 plus
```

jaleszek:
to fix

PR #368, opened 2015-04-16
Fix the fk for the virus scanner

jaleszek:


matthewford:
I would prefer if we could just make this a polymorphic association

jaleszek:
me 2 but backlog is so long ..

PR #373, opened 2015-04-16
Make review functionality more closed.

jaleszek:
Found that Assessor is able to review the all User forms even if is not
assigned to that case and is not a lead either. Added some restrictions to
reviewing the resources also as It was designed initially to use for the Admin,
but in case of Assessors we should be more suspicious.

### DIFF
--- a/app/controllers/account/collaborators_controller.rb
+++ b/app/controllers/account/collaborators_controller.rb
@@ -1,6 +1,7 @@
 class Account::CollaboratorsController < Account::BaseController
-  
+
   before_action :require_to_be_not_current_user!, only: [:destroy]
+  before_action :restrict_access_if_admin_in_read_only_mode!
 
   before_action :set_form_answer
 
@@ -15,8 +16,8 @@ class Account::CollaboratorsController < Account::BaseController
   end
   expose(:add_collaborator_interactor) do
     AddCollaborator.new(
-        current_user, 
-        account, 
+        current_user,
+        account,
         {})
   end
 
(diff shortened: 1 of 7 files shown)


app/controllers/account/collaborators_controller.rb
```diff
@@ -1,6 +1,7 @@
 class Account::CollaboratorsController < Account::BaseController
-  
+
```
matthewford: Extra empty line detected at class body beginning.

app/controllers/account/collaborators_controller.rb
```diff
@@ -15,8 +16,8 @@ class Account::CollaboratorsController < Account::BaseController
   end
   expose(:add_collaborator_interactor) do
     AddCollaborator.new(
-        current_user, 
-        account, 
+        current_user,
```
matthewford: Indent the first parameter one step more than the previous line.

app/controllers/application_controller.rb
```diff
@@ -31,19 +31,38 @@ def after_sign_in_path_for(resource)
   end
 
   def admin_in_read_only_mode?
-    @admin_in_read_only_mode ||= admin_signed_in? &&
+    @admin_in_read_only_mode ||= (admin_signed_in? || assessor_signed_in?) &&
                                  session["warden.user.user.key"] &&
                                  session[:admin_in_read_only_mode]
+
```
matthewford: Extra empty line detected at method body end.

app/controllers/content_only_controller.rb
```diff
@@ -44,6 +44,8 @@ class ContentOnlyController < ApplicationController
                   :privacy,
                   :cookies
                 ]
+  before_action :restrict_access_if_admin_in_read_only_mode!,
+                only: [ :dashboard ]
```
matthewford: Space inside square brackets detected.

PR #375, opened 2015-04-16
Fix review functionality.

jaleszek:
Found that Assessor is able to review the all User forms even if is not
assigned to that case and is not a lead either. Added some restrictions to
reviewing the resources also as It was designed initially to use for the Admin,
but in case of Assessors we should be more suspicious.

### DIFF
--- a/app/controllers/account/collaborators_controller.rb
+++ b/app/controllers/account/collaborators_controller.rb
@@ -1,6 +1,7 @@
 class Account::CollaboratorsController < Account::BaseController
-  
+
   before_action :require_to_be_not_current_user!, only: [:destroy]
+  before_action :restrict_access_if_admin_in_read_only_mode!
 
   before_action :set_form_answer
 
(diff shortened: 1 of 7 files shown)


app/controllers/account/collaborators_controller.rb
```diff
@@ -1,6 +1,7 @@
 class Account::CollaboratorsController < Account::BaseController
-  
+
```
matthewford: Extra empty line detected at class body beginning.

app/controllers/account/collaborators_controller.rb
```diff
@@ -15,8 +16,8 @@ class Account::CollaboratorsController < Account::BaseController
   end
   expose(:add_collaborator_interactor) do
     AddCollaborator.new(
-        current_user, 
-        account, 
+        current_user,
```
matthewford: Indent the first parameter one step more than the previous line.

RuslanHamidullin:
Good to merge :+1:  Waiting travis

PR #381, opened 2015-04-17
remove couple of questions from sus.dev. form (B10, C6.5, C6.6)

dreamfall:


app/forms/qae_2014_forms/sustainable_development/sustainable_development_step2.rb
```diff
@@ -164,7 +165,7 @@ def development_step2
           required
           context %(
             <p>
-              'Leadership dimensions' means actively promoting effective, participative systems of governance in all levels of society 
+              'Leadership dimensions' means actively promoting effective, participative systems of governance in all levels of society
```
matthewford: Line is too long. [134/100]

jaleszek:
Do we care about removing these answers from the db for already stored records ?

dreamfall:
@jaleszek I guess no, there no real applications now, so it's ok to have some excess data in couple of test records

PR #396, opened 2015-04-19
remove-link fix for supporters

dreamfall:


app/assets/stylesheets/frontend/views/award_form.scss
```diff
@@ -150,6 +150,12 @@ label > .visible-read-only,
   }
 }
 
+.supporters-list > .read-only {
+  .remove-link {
+    display: inline !important;
```
matthewford: !important should not be used

PR #404, opened 2015-04-20
use strict 15 limit, add 10% buffer if limit is > 15

dreamfall:
fix step validation

### DIFF
--- a/lib/qae_form_builder/textarea_question.rb
+++ b/lib/qae_form_builder/textarea_question.rb
@@ -4,7 +4,11 @@ class QAEFormBuilder
       result = super
 
       limit = question.delegate_obj.words_max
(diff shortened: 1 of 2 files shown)


lib/qae_form_builder/textarea_question.rb
```diff
@@ -4,7 +4,11 @@ def errors
       result = super
 
       limit = question.delegate_obj.words_max
-      if limit && question.input_value && question.input_value.split(" ").length > limit
+
+      limit_with_buffer = limit
+      limit_with_buffer = (limit + limit * 0.1).to_i + 1 if limit > 15
+
+      if limit_with_buffer && question.input_value && question.input_value.split(" ").length > limit_with_buffer
```
matthewford: Line is too long. [112/100]

PR #405, opened 2015-04-20
Pdf fixes

RuslanHamidullin:
* hidden questions now should be visible
* Calculated financial data shouldn't be decimal
* queen_award_holder_details: fixed missing category


app/pdf_generators/qae_pdf_forms/custom_questions/by_year.rb
```diff
@@ -33,7 +33,7 @@ def render_years_labels_table
 
   def render_years_table
     rows = if CALCULATED_FINANCIAL_DATA.include?(question.key)
-      get_audit_data(question.key)
+      get_audit_data(question.key).map(&:to_i).map(&:to_s)
```
matthewford: Use 2 (not -5) spaces for indentation.

PR #413, opened 2015-04-21
Set up the date format in import

jaleszek:


app/models/users_import/builder.rb
```diff
@@ -23,7 +23,7 @@ def process
         u.agreed_with_privacy_policy = "1"
         u.skip_confirmation!
         if u.save
-          u.update_column(:created_at, user["UserCreationDate"])
+          u.update_column(:created_at, Date.strptime(user["UserCreationDate"], "%d/%m/%y")) if user["UserCreationDate"].present?
```
matthewford: Line is too long. [128/100]

PR #418, opened 2015-04-22
Remove the reset expiration time for imported users.

jaleszek:
Users migrated from another app shouldn't have expiration of their reset
password tokens. Assumed that all tokens sent on release date are these sent with import.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -98,6 +98,15 @@ class User < ActiveRecord::Base
     "#{first_name} #{last_name}".strip
   end
 
+  def reset_password_period_valid?
+    release_date = Date.new(2015, 4, 21)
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -98,6 +98,15 @@ def full_name
     "#{first_name} #{last_name}".strip
   end
 
+  def reset_password_period_valid?
+    release_date = Date.new(2015, 4, 21)
+    if imported? && reset_password_sent_at.present? && reset_password_sent_at.to_date == release_date
```
matthewford: Line is too long. [101/100]

RuslanHamidullin:
Good. Waiting travis

PR #458, opened 2015-05-04
[QAE1178 | CRON STUFF]

RuslanHamidullin:
* Implemented CronJob class that does the locking stuff and adds a small sleep say 3 secs before the work
* Skylight: disabled alert messages on dev / staging envs
* Cron / Capistrano: fixed issue with binstubs messages

Changed the schedule for some tasks:

`FormAnswerStateMachine.trigger_deadlines` only needs to be run the day after the deadline to move form states, so run at 12:30am.

`Notifiers::EmailNotificationService.run` is not time sensitive so just run once an hour.

### DIFF
--- a/app/models/cron_job.rb
+++ b/app/models/cron_job.rb
@@ -1,0 +1,8 @@
+class CronJobRun
+  def self.run(lock_name, timeout_seconds=0, sleep_before_run=3, &block)
+    User.with_advisory_lock(lock_name, timeout_seconds) do
+      sleep(sleep_before_run) if sleep_before_run.to_i > 0
+      block.call
+    end
+  end
+end
--- a/app/models/cron_job.rb
+++ b/app/models/cron_job.rb
@@ -1,5 +1,5 @@
 class CronJobRun
-  def self.run(lock_name, timeout_seconds=0, sleep_before_run=3, &block)
+  def self.run(lock_name, timeout_seconds=0, sleep_before_run=5, &block)
     User.with_advisory_lock(lock_name, timeout_seconds) do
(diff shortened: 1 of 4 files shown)


app/models/cron_job.rb
```diff
@@ -0,0 +1,8 @@
+class CronJobRun
+  def self.run(lock_name, timeout_seconds=0, sleep_before_run=3, &block)
```
matthewford: Surrounding space missing in default value assignment.

config/schedule.rb
```diff
@@ -6,10 +9,10 @@
 
 # Do not run jobs with the same offset
 
-every 7.minutes do
-  runner "User.with_advisory_lock('email_notification', 0) { Notifiers::EmailNotificationService.run }"
+every :day, at: "12:30am" do
+  runner "CronJobRun.run('deadlines', 0, 5) { FormAnswerStateMachine.trigger_deadlines }"
```
matthewford: if `sleep_before_run` is 5, perhaps that should be the default?

matthewford: if we make the sensible defaults - can we just reduce this to 

`CronJobRun.run('deadlines') { FormAnswerStateMachine.trigger_deadlines }``

RuslanHamidullin:
```
sleep_before_run=3
```
is default, which would be overriden by passing custom

RuslanHamidullin: ahh ,agree

RuslanHamidullin: fixing

RuslanHamidullin:
Fixed )

PR #460, opened 2015-05-04
[Whenever] separated env lists

RuslanHamidullin:
Running of cron tasks based on ENV

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -10,9 +10,16 @@ set :output, File.expand_path("#{File.dirname __FILE__}/../log/cron_log.log")
 # Do not run jobs with the same offset
 
 every :day, at: "12:30am" do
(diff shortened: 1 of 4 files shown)


config/schedule.rb
```diff
@@ -10,9 +10,16 @@
...
+  runner "CronJob.run('deadlines') { FormAnswerStateMachine.trigger_deadlines }"
 end
 
 every :hour do
-  runner "CronJobRun.run('email_notification') { Notifiers::EmailNotificationService.run }"
+  runner "CronJob.run('email_notification') { Notifiers::EmailNotificationService.run }"
+end
+
+case @environment
+when 'production'
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

RuslanHamidullin:
Need to fix

PR #461, opened 2015-05-04
[Whenever ] cron separated envs

RuslanHamidullin:
Running of cron tasks based on ENV

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -10,11 +10,18 @@ set :output, File.expand_path("#{File.dirname __FILE__}/../log/cron_log.log")
 # Do not run jobs with the same offset
 
 every :day, at: "12:30am" do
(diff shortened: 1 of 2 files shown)


config/schedule.rb
```diff
@@ -10,13 +10,16 @@
...
 
 every :hour do
-  runner "CronJobRun.run('email_notification') { Notifiers::EmailNotificationService.run }"
+  runner "CronJob.run('email_notification') { Notifiers::EmailNotificationService.run }"
 end
 
-every :sunday, at: "12pm" do
-  runner "PerformancePlatformService.run"
+case @environment
+when 'production'
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

PR #467, opened 2015-05-05
date can be invalid

dreamfall:


app/models/reports/form_answer.rb
```diff
@@ -390,8 +390,9 @@ def date_started_trading
     day = doc("started_trading_day")
     month = doc("started_trading_month")
     year = doc("started_trading_year")
+
     if year && month && day
-      Date.new(year.to_i, month.to_i, day.to_i).strftime("%m/%d/%Y")
+      Date.new(year.to_i, month.to_i, day.to_i).strftime("%m/%d/%Y") rescue nil
```
matthewford: Avoid using `rescue` in its modifier form.

RuslanHamidullin:
True :+1:

PR #483, opened 2015-05-07
do not use flash messages when redirected to root url

dreamfall:


app/controllers/users/press_summaries_controller.rb
```diff
@@ -1,8 +1,8 @@
 class Users::PressSummariesController < Users::BaseController
   skip_before_action :authenticate_user!
 
-  before_action :check_deadline, :load_press_summary
-  before_action :check_promotion_award_acceptance, except: [:acceptance, :update_acceptance]
+  before_action :check_deadline, :load_press_summary, except: [:success, :failure]
+  before_action :check_promotion_award_acceptance, except: [:acceptance, :update_acceptance, :success, :failure]
```
matthewford: Line is too long. [112/100]

app/controllers/users/support_letters_controller.rb
```diff
@@ -37,17 +37,17 @@ def show
...
-      )
-    end
+  def support_letter_params
+    params.require(:support_letter).permit(
+      :first_name,
+      :last_name,
+      :relationship_to_nominee,
+      :attachment
+    )
+   end
```
matthewford: `end` at 52, 3 is not aligned with `def` at 45, 2

PR #485, opened 2015-05-08
[Rack Mini Profiler] let's disable it by default

RuslanHamidullin:
@matthewford @dreamfall @hammerfunk  how about disabling of rack_mini_profiler by default.
It adds some overhead, which make debugging of JS in firebug or chrome console harder.
Also, we use do not use it often.
My suggestion is to make it customizable and disabled by default.
Your opinion guys?


dreamfall:
:+1:

RuslanHamidullin:
:bomb:

Issue #364, opened 2015-04-16
DEV TASK: replace Date.today.year with AwardYear.current.year


RuslanHamidullin:
Need to replace Date.today.year with AwardYear.current.year in couple of places

RuslanHamidullin:
These are the found dynamic dates (can be not all)
https://docs.google.com/a/bitzesty.com/spreadsheets/d/12gZm1_0iFWWxN6ytOittvCiEHymW2cQr1WG_Durqf7U/edit#gid=0

matthewford:
@RuslanHamidullin can we close this now?

RuslanHamidullin:
sec, checking

RuslanHamidullin:
@matthewford nope, I still see some places which probably (need to clarify) need to correct:
```
/home/alkapone/projects/qae/app/form_pointers/financial_year_pointer.rb:
   69      month = "0" + month if month.size == 1
   70  
   71:     [day, month, Date.today.year ]
   72    end
   73  end

/home/alkapone/projects/qae/app/models/account.rb:
   22  
   23    def has_trade_award_in_this_year?
   24:     form_answers.for_year(Date.today.year + 1).
   25                   for_award_type(:trade).
   26                   present?

/home/alkapone/projects/qae/app/pdf_generators/qae_pdf_forms/custom_questions/by_year.rb:
  103      month = "0" + month if month.size == 1
  104  
  105:     [day, month, Date.today.year ]
  106    end
  107  end

/home/alkapone/projects/qae/lib/qae_form_builder/innovation_financial_year_date_question.rb:
    9        end
   10  
   11:       date << Date.today.year.to_s
   12  
   13        date = Date.parse(date.join("/")) rescue nil

/home/alkapone/projects/qae/spec/models/account_spec.rb:
    4    let(:user) { FactoryGirl.build(:user) }
    5    let(:account) { user.account }
    6:   let(:current_year) { AwardYear.where(year: Date.today.year + 1).first_or_create }
    7    let(:previous_year) { AwardYear.where(year: current_year).first_or_create }
    8  

/home/alkapone/projects/qae/spec/models/form_answer_spec.rb:
   24    context "URN" do
   25      before do
   26:       FormAnswer.connection.execute("ALTER SEQUENCE urn_seq_#{Date.today.year + 1} RESTART")
   27:       FormAnswer.connection.execute("ALTER SEQUENCE urn_seq_promotion_#{Date.today.year + 1} RESTART")
   28      end
   29  
[...]

PR #490, opened 2015-05-08
[Hardcoded years] Fixes

RuslanHamidullin:
Fixes of issue https://github.com/bitzesty/qae/issues/364

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -21,7 +21,7 @@ class Account < ActiveRecord::Base
   end
 
   def has_trade_award_in_this_year?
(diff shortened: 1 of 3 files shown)


app/models/account.rb
```diff
@@ -21,7 +21,7 @@ def collaborators_without(user)
   end
 
   def has_trade_award_in_this_year?
-    form_answers.for_year(Date.today.year + 1).
+    form_answers.for_year(AwardYear.current.year.to_s).
```
matthewford: Place the . on the next line, together with the method name.

PR #491, opened 2015-05-08
fix randomly failing csv test

dreamfall:


app/models/reports/all_entries.rb
```diff
@@ -171,9 +171,11 @@ def initialize(year)
 
   def build
     rows = []
-    scope = ::FormAnswer.select(:id).where(award_year_id: @year.id)
+
+    scope = @year.form_answers.order(:id)
     scope.find_in_batches do |batch|
       form_answers = FormAnswer.where(id: batch.map(&:id))
```
RuslanHamidullin: Probably we can make it bit faster via fetching only id columns:
```
scope = @year.form_answers.order(:id).pluck(:id)
 scope.find_in_batches do |batch_ids|
       form_answers = FormAnswer.where(id: batch_ids)
```

app/models/reports/cases_status_report.rb
```diff
@@ -98,9 +98,11 @@ def initialize(year)
 
   def build
     rows = []
-    scope = ::FormAnswer.select(:id).where(award_year_id: @year.id, submitted: true)
+
+    scope = @year.form_answers.where(submitted: true).order(:id)
```
RuslanHamidullin: same like above ^

app/models/reports/registered_users.rb
```diff
@@ -111,9 +111,10 @@ def initialize(year)
   def build
     rows = []
 
-    scope = ::FormAnswer.select(:id).where(award_year_id: @year.id)
+    scope = @year.form_answers.order(:id)
```
RuslanHamidullin: same like above ^

RuslanHamidullin:
Good :+1:

Issue #358, opened 2015-04-16
Clean up orphaned files which are removed from form.


RuslanHamidullin:
@matthewford 
Currently we do not remove form attachment files (stored in FormAnswerAttachment) if it removed from Form.
It explained as we need to click "Save and Continue" on JS form in order to apply removing attachment from Form document hash.

Looks like we need to remove unused FormAnswerAttachments's via for example cron task?

matthewford:
Think we need to discuss this.

RuslanHamidullin:
@matthewford gonna take a look to this one.
So possible ways to solve:
1) Add cron script, which would clean up Attachments related to forms, but not saved in document of form (probably weekly)
2) Add ajax removing of attachments on click "Remove"

matthewford:
think step 2) would be preferable, although when you say ajax I assume you mean a way with a non-js fallback?

RuslanHamidullin:
In non - js version already implemented removing of attachment

RuslanHamidullin:
Ok, gonna add removing of attachment via AJAX for JS version

PR #494, opened 2015-05-11
[Form Attachments] AJAX removing of attachment record

RuslanHamidullin:
[Form Attachments] AJAX removing of attachment record.
Fixes https://github.com/bitzesty/qae/issues/358


RuslanHamidullin:
need to fix for Organisational Charts too

PR #495, opened 2015-05-11
Clean up orphaned files which are removed from form

RuslanHamidullin:
Fixes https://github.com/bitzesty/qae/issues/358
1) Form Attachments: AJAX removing of attachment record 
2) Form | Organisational Chart Upload | Remove: implemented AJAX removing of file
3) Form | Organisational Chart Upload:  added missing max attachments validations for Trade and Dev forms

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -310,8 +310,20 @@ jQuery ->
       $(li).find("textarea").each process_input
       idx++
 
-  appendLinkDocumentRemoveLink = (div) ->
+  appendRemoveLinkForWebsiteLink = (div) ->
     remove_link = $("<a>").addClass("remove-link").prop("href", "#").text("Remove")
+    div.append(remove_link)
+
+  appendRemoveLinkForAttachment = (div, data) ->
+    attachment_id = data.result['id']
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/application.js.coffee
```diff
@@ -310,10 +310,27 @@ jQuery ->
...
-  appendLinkDocumentRemoveLink = (div) ->
+  appendRemoveLinkForWebsiteLink = (div) ->
     remove_link = $("<a>").addClass("remove-link").prop("href", "#").text("Remove")
     div.append(remove_link)
 
+  appendRemoveLinkForAttachment = (div, name, data) ->
+    attachment_id = data.result['id']
+    form_answer_id = data.result['form_answer_id']
+
+    destroy_url = if name == "org_chart"
```
matthewford: would it be possible to include this info in a data-attribute that we just read? rather than hard coding this?

RuslanHamidullin: @matthewford re: "would it be possible to include this info in a data-attribute that we just read? rather than hard coding this?"
 => agree, will do

RuslanHamidullin: Fixed code review comment via https://github.com/bitzesty/qae/pull/498

PR #500, opened 2015-05-12
Tweaking layout on sic code and previous winning sections

mauricioac:
This PR fixes position of save button on sic code editing form and also fixes the layout of the previous winning section, also editing form.

It also checks for a variable on the `handleCompanyDetailsForm`, which was causing error by being undefined and trying to perform a `replace`.

### DIFF
--- a/app/assets/stylesheets/admin/page-applications.scss
+++ b/app/assets/stylesheets/admin/page-applications.scss
@@ -462,3 +462,21 @@
     clear:both;
   }
 }
+
+.form-group.previous-wins-form.form-edit {
+  .form-fields {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/admin/page-applications.scss
```diff
@@ -461,4 +461,22 @@
   .lte-ie8 & {
     clear:both;
   }
+}
+
+.form-group.previous-wins-form.form-edit {
+  .form-fields {
+    display: block !important;
```
matthewford: !important should not be used

PR #501, opened 2015-05-12
Fix missing audit certificate link

jaleszek:


app/views/admin/form_answers/financial_summary/_audit_certificate.html.slim
```diff
@@ -1,5 +1,7 @@
 label.form-label Audit Certificate
-p = link_to "View Audit Certificate", "#", target: "_blank"
+- if resource.audit_certificate.present?
```
RuslanHamidullin: @jaleszek need a virus check for this one, as it's user's upload.
We have special policy for this case https://github.com/bitzesty/qae/blob/master/app/policies/form_answer_policy.rb#L43
So would be:
```
- if policy(resource).download_audit_certificate_pdf?
  = link_to "View Audit Certificate", resource.audit_certificate.attachment_url,
                                                       target: "_blank", class: "js-audit-certificate-title"
```

PR #502, opened 2015-05-12
Post submission submitted application

dreamfall:


app/decorators/form_answer_decorator.rb
```diff
@@ -107,6 +107,10 @@ def sic_code_name
     end
   end
 
+  def application_in_progress?
```
jaleszek: there is another state `application_in_progress` I hope it won't be misleading

dreamfall: @jaleszek you're right, I should change this method's name

app/views/content_only/post_submission/_submitted.html.slim
```diff
@@ -0,0 +1,19 @@
+- applications = FormAnswerDecorator.decorate_collection(award_applications.where(state: %w(assessment_in_progress submitted not_submitted application_in_progress)))
```
jaleszek: Technically we shouldn't have to look for a `submitted` and `application_in_progress` here as these states can't exist after the submission deadline.

jaleszek: in 24hours after the submission deadline so in fact for sure after the shortlisting

PR #508, opened 2015-05-14
Validate the password change for the Admin

jaleszek:
https://trello.com/c/dPPZVExy/27-as-an-admin-when-i-edit-a-user-and-type-different-password-for-password-confirmation-it-shows-that-successfully-edited-user


app/models/user.rb
```diff
@@ -25,6 +25,7 @@ def after_initialize
   validates :last_name, presence: true, if: :first_step?
   validates :job_title, presence: true, if: :first_step?
   validates :phone_number, presence: true, if: :first_step?
+  validates_confirmation_of :password
```
matthewford: Prefer the new style validations `validates :column, confirmation: value` over `validates_confirmation_of`.

PR #510, opened 2015-05-14
Security headers

RuslanHamidullin:
This is fix for "[Report Only] Refused to load the font ..." error message on Chrome/FF console
By default, we do not specify a font-src which means the value should have taken the value of default-src ('none' in this case)

Below we add "font-src data:;" to whitelist the font being loaded.
```
  config.csp = {
    default_src: "https: inline eval",
    img_src: "https:",
    font_src: "'self' data:" # font_src - defines valid sources of fonts. (http://content-security-policy.com/)
  }
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,9 +6,13 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   # TODO: remove back once go live
-  # unless (Rails.env.test? || Rails.env.development?)
-  #   ensure_security_headers
-  # end
+  unless (Rails.env.test? || Rails.env.development?)
+    # Checking on ENV["AWS_ACCESS_KEY_ID"] exists
+    # In order to avoid exceptions on dev and demo servers too
+    # as they also use production env
+
+    ensure_security_headers if ENV["AWS_ACCESS_KEY_ID"]
+  end
 
   before_filter :configure_permitted_parameters, if: :devise_controller?
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,7 +5,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  ensure_security_headers if ENV['ENSURE_SECURITY_HEADERS']
+  ensure_security_headers if ENV["ENSURE_SECURITY_HEADERS"]
 
   before_filter :configure_permitted_parameters, if: :devise_controller?
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,14 +5,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  # TODO: remove back once go live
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -6,9 +6,13 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   # TODO: remove back once go live
-  # unless (Rails.env.test? || Rails.env.development?)
-  #   ensure_security_headers
-  # end
+  unless (Rails.env.test? || Rails.env.development?)
```
matthewford: Don't use parentheses around the condition of an `unless`.

RuslanHamidullin:
@matthewford can you please take a look ^

app/controllers/application_controller.rb
matthewford: could we make this just `if Rails.env.production`?

app/controllers/application_controller.rb
```diff
@@ -6,9 +6,13 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
 
   # TODO: remove back once go live
```
matthewford: remove comments

app/controllers/application_controller.rb
```diff
@@ -6,9 +6,13 @@ class ApplicationController < ActionController::Base
...
-  # unless (Rails.env.test? || Rails.env.development?)
-  #   ensure_security_headers
-  # end
+  unless (Rails.env.test? || Rails.env.development?)
+    # Checking on ENV["AWS_ACCESS_KEY_ID"] exists
+    # In order to avoid exceptions on dev and demo servers too
+    # as they also use production env
+
+    ensure_security_headers if ENV["AWS_ACCESS_KEY_ID"]
```
matthewford: I would prefer a different key, perhaps we need a new env variable?

RuslanHamidullin:
@matthewford agree, gonna correct

app/controllers/application_controller.rb
```diff
@@ -5,10 +5,7 @@ class ApplicationController < ActionController::Base
   # For APIs, you may want to use :null_session instead.
   protect_from_forgery with: :exception
 
-  # TODO: remove back once go live
-  # unless (Rails.env.test? || Rails.env.development?)
-  #   ensure_security_headers
-  # end
+  ensure_security_headers if ENV['ENSURE_SECURITY_HEADERS']
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

RuslanHamidullin:
Need to fix fileupload issue at first:
fileupload_issue

matthewford:
Yeah seems its the iframe upload issue - but its the same domain so in theory should be fine...

RuslanHamidullin:
yeah,
I replaced:
```
config.csp = {
    default_src: "https: inline eval",
```
with:
```
config.csp = {
    default_src: "https: self"
```
Probably should fix. Gonna test

RuslanHamidullin:
Nope 
```
config.csp = {
    default_src: "https: self"
```
Causes:
nope

Gonna debug

RuslanHamidullin:
Alright, fixed iframe issue - now files are uploading.
But still some warnings. DEbugging
upload_issues

PR #525, opened 2015-05-18
Admin dashboard needs a year filter

jaleszek:


jaleszek:
@dreamfall can you take a look please ?

PR #526, opened 2015-05-18
Admin textarea size fix, cancel buttons and fixed table overflow

hammerfunk:


app/assets/stylesheets/admin/page-applications.scss
```diff
@@ -142,6 +142,10 @@
       .form-save-link,
       .form-cancel-link {
         display: inline-block;
+
+        &.form-block {
```
matthewford: Nesting should be no greater than 4, but was 5

PR #527, opened 2015-05-18
Restyled 2fa pages

hammerfunk:


app/assets/javascripts/twofactor.js.coffee
```diff
@@ -2,3 +2,7 @@
 #= require jquery_ujs
 #= require bootstrap-sprockets
 #= require devise_authy
+
+$(document).ready(() ->
```
matthewford: Empty parameter list is forbidden

app/views/devise/devise_authy/verify_authy.html.slim
```diff
@@ -0,0 +1,32 @@
...
+  br
+
+  = submit_tag I18n.t('submit_token', {:scope => 'devise'}), :class => 'btn btn-primary'
+
+  br
+
+  small
+    = link_to '?', 'https://authy.zendesk.com/hc/en-us/requests/new', :id => 'authy-help', target: "_blank"
+    / br
+    / = authy_request_sms_link
```
RuslanHamidullin: Looks like:
```
authy_request_sms_link
```
should be, does it?

app/assets/javascripts/twofactor.js.coffee
```diff
@@ -2,3 +2,7 @@
 #= require jquery_ujs
 #= require bootstrap-sprockets
 #= require devise_authy
+
+$(document).ready( ->
+  $("html").removeClass("no-js").addClass("js")
+)
```
RuslanHamidullin: So, our application.js.coffee looks like requiring whole tree
https://github.com/bitzesty/qae/blob/master/app/assets/javascripts/application.js.coffee#L16
```
#= require_tree .
```
not 100% sure, but looks like this one:
```
 $("html").removeClass("no-js").addClass("js")
```
would be called for user forms too. Does it?

app/views/devise/devise_authy/verify_authy_installation.html.slim
```diff
@@ -0,0 +1,24 @@
...
+  br
+
+  = submit_tag I18n.t('enable_my_account', {:scope => 'devise'}), :class => 'btn btn-primary'
+
+  br
+
+  small
+    = link_to "Help", 'https://authy.zendesk.com/hc/en-us/requests/new', :id => 'authy-help', target: "_blank"
+    / br
+    / = authy_request_sms_link
```
RuslanHamidullin: Same note with missing authy_request_sms_link like above ^

app/views/devise/devise_authy/verify_authy.html.slim
hammerfunk: So I added it back but it just creates a link to a broken page. I don't know if we've added this as a setting. Here's the error:
`No route matches [GET] "/admins/request-sms"`

app/assets/javascripts/twofactor.js.coffee
hammerfunk: So the 2fa page has it's own layout `twofactor.html.slim` and it uses it's own main JS `twofactor.js.coffee` that only calls a few things, but it doesn't call `#= require_tree .`

app/assets/javascripts/application.js.coffee
```diff
@@ -13,9 +14,11 @@
 #= require browser-check
 #= require vendor/zxcvbn
 #= require vendor/jquery-debounce
-#= require_tree .
+#= require_tree ./frontend
 
 jQuery ->
+  $("html").removeClass("no-js").addClass("js")
```
RuslanHamidullin: @hammerfunk just to confirm:
Do we need:
````
 jQuery ->
+  $("html").removeClass("no-js").addClass("js")
```
in app/assets/javascripts/application.js.coffee too?

hammerfunk: Yes we do, otherwise it will default to non-js views since all the js only CSS will not activate.

RuslanHamidullin:
Good to merge :+1:

PR #539, opened 2015-05-20
Scheduled emails issue

jaleszek:
- trigger changes transition immediately after the deadline edition

### DIFF
--- a/app/workers/submission_deadline_states_transition_worker.rb
+++ b/app/workers/submission_deadline_states_transition_worker.rb
@@ -1,0 +1,9 @@
+class SubmissionDeadlineStatesTransitionWorker
+  include Shoryuken::Worker
+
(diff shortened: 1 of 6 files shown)


app/workers/submission_deadline_states_transition_worker.rb
```diff
@@ -0,0 +1,9 @@
+class SubmissionDeadlineStatesTransitionWorker
+  include Shoryuken::Worker
+
+  shoryuken_options queue: "#{Rails.env}_default", auto_delete: true
+
+  def perform(sqs_msg)
```
matthewford: Unused method argument - `sqs_msg`. If it's necessary, use `_` or `_sqs_msg` as an argument name to indicate that it won't be used. You can also write as `perform(*)` if you want the method to accept any arguments but don't care about them.

RuslanHamidullin:
Well done :+1:

PR #555, opened 2015-05-26
BZSTAGING | No need to ask to reboot server on errors while deploy

RuslanHamidullin:
THIS applying only for BZSTAGING!
So we are often getting random "Allocate Memory" issues on deploy (a specially after number of deploys).
So solution is to reboot server after each deploy in order to reduce possible "Allocate memory" issues.

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -13,15 +13,11 @@ set :slack_icon_emoji,   ->{ ":rocket:" }
 set :slack_channel,      ->{ "#qae" }
 set :migration_role, 'app'
 
-set :stages, %w(production staging dev demo)
-set :default_stage, 'staging'
+set :stages, %w(production staging bzstaging)
+set :default_stage, 'bzstaging'
 set :use_sudo, false
 
-if ENV["OLD_SERVERS"]
-  set :deploy_to, "/home/qae/application"
-else
-  set :deploy_to, "/home/#{fetch(:user)}/#{fetch(:application)}"
-end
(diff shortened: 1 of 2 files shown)


config/deploy.rb
```diff
@@ -13,15 +13,11 @@
 set :slack_channel,      ->{ "#qae" }
 set :migration_role, 'app'
 
-set :stages, %w(production staging dev demo)
-set :default_stage, 'staging'
+set :stages, %w(production staging bzstaging)
+set :default_stage, 'bzstaging'
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

config/deploy.rb
```diff
@@ -60,6 +56,14 @@
     end
   end
 
+  # Running only on dev server (as it has small CPU / Memory resources)
+  desc 'Reboot Server'
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

config/deploy/bzstaging.rb
```diff
@@ -2,4 +2,4 @@
 set :rails_env, 'bzstaging'
 set :branch, ENV["BRANCH"] || 'master'
 
-server '162.13.46.189', user: 'qae', roles: %w{web app db}
+server '162.13.46.189', user: 'qae', roles: %w{web app db bzstaging}
```
matthewford: Prefer double-quoted strings unless you need single quotes to avoid extra backslashes for escaping.

RuslanHamidullin:
:bomb:

Issue #552, opened 2015-05-26
don't know how to replicate it but sometimes I get this error when trying to download the pdfs on admin dashboard `uninitialized constant Reports::FormAnswer::AWARD_TYPE_FULL_NAMES`


RuslanHamidullin:
don't know how to replicate it but sometimes I get this error when trying to download the pdfs on admin dashboard
```
uninitialized constant Reports::FormAnswer::AWARD_TYPE_FULL_NAMES
```
![screen_shot_2015-05-26_at_11 24 10_1024](https://cloud.githubusercontent.com/assets/358508/7810456/de258ea6-03aa-11e5-93ee-a89fe3c6e56f.png)

jaleszek:
I believe that fixed by: https://github.com/bitzesty/qae/commit/a5c827145ed910f9f22ef0eb2bfcbb21e4ebac9b

PR #564, opened 2015-06-05
fix constant namespace

dreamfall:
https://github.com/bitzesty/qae/issues/552

it was already fixed on line 42, but found another constant usage on line 47


RuslanHamidullin:
Good :+1:

PR #570, opened 2015-06-19
add cumulative count to performance platform

dreamfall:
And transactions by channel only include submitted forms

### DIFF
--- a/app/services/performance_platform_service.rb
+++ b/app/services/performance_platform_service.rb
@@ -14,12 +14,12 @@ class PerformancePlatformService
   }
 
   POSSIBLE_RANGES = [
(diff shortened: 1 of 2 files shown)


app/services/performance_platform_service.rb
```diff
@@ -14,12 +14,12 @@ class PerformancePlatformService
   }
 
   POSSIBLE_RANGES = [
-   "0-percent",
-   "1-24-percent",
-   "25-49-percent",
-   "50-74-percent",
+   "100-percent",
```
matthewford: Use 2 spaces for indentation in an array, relative to the start of the line where the left bracket is.

PR #576, opened 2015-07-23
fix character counter

dreamfall:


app/assets/javascripts/frontend/text-character-count.js
```diff
@@ -79,6 +79,7 @@ $.fn.charcount = function() {
         lastIndex = oldText.lastIndexOf("\n") + 1;
       }
       newText = oldText.substr(0, lastIndex) + " ";
+      newText = newText.replace(/\s+$/g, '');
```
matthewford: Mixed double and single quotes.

PR #583, opened 2015-08-26
Fix for the checkbox filters on tables

mauricioac:
This PR adds a timeout before applying changes to filters when clicking a checkbox. Before it had no timeout, so some browsers weren't able to process the filter fast enough.


rusllonrails:
Looks good man :+1:

PR #590, opened 2015-09-10
Assessors can see a read only admin comments and both see the flags a…

hammerfunk:
Assessors can see a read only admin comments
Both see the both flags count and total comments count on the list page


rusllonrails:
Looks fine

PR #595, opened 2015-09-16
Validate moderated appraisals

hammerfunk:


app/forms/appraisal_form.rb
```diff
@@ -212,16 +212,22 @@ def self.desc(key)
...
+    if moderated
+      assessment_types = [:verdict]
+    else
+      assessment_types = [:rag, :non_rag, :verdict]
+    end
     const_get(award_type.upcase).map do |k, obj|
-      methods = Array(rate(k))
-      methods << desc(k) if [:rag, :non_rag, :verdict].include?(obj[:type])
+      methods = Array.new
+      methods << Array(rate(k)) if ((!moderated) || (moderated && obj[:type] == :verdict))
```
rusllonrails: Also you can do same via:
```
methods << Array(rate(k)) if !moderated || moderated && obj[:type] == :verdict
```
instead of:
```
methods << Array(rate(k)) if ((!moderated) || (moderated && obj[:type] == :verdict))
```
Do not need to correct now. Just fyi :smile:

rusllonrails:
Good :+1:

PR #596, opened 2015-09-16
Fixes for org chart uploads

mauricioac:
This PR adds code to help on the org chart upload issue.
The issue is a race condition that happens between saving the file and the form with the reference to the file.
So in this scenario, the file is there, but lost, no one referenciing it.

This branch insted of triggering the autosave function, actually autosaves immediately after getting success response from backend, making browsers wait till saved before the user try to navigate out of the form.

This branch also removes the validation on org chart to allow multiple files to exist (the frontend doesn't allow more uploads though).

### DIFF
--- a/app/models/form_answer_attachment.rb
+++ b/app/models/form_answer_attachment.rb
@@ -62,10 +62,10 @@ class FormAnswerAttachment < ActiveRecord::Base
     # possible improvements:
     # - validating all upload questions in this way
     # - smart collecting max-attachments information from the form structure
-    if question_key == "org_chart" && new_record?
-      if self.class.where(question_key: question_key, form_answer_id: form_answer_id).count > 0
-        errors.add(:base, "Maximum amount of these kind of files reached.")
-      end
-    end
+    # if question_key == "org_chart" && new_record?
+    #   if self.class.where(question_key: question_key, form_answer_id: form_answer_id).count > 0
+    #     errors.add(:base, "Maximum amount of these kind of files reached.")
+    #   end
+    # end
   end
 end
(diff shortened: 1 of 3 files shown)


matthewford:
not sure why the schema keeps updating here @mauricioac can you drop db rerun the migrations?

mauricioac:
will do

mauricioac:
@matthewford dropped the database, recreated and ran migrations. No changes from this commit. I guess the schema wasn't updated on master when I branched. Do you want me to remove the file from the commit?

eritiro:
@matthewford need to check but probably @mauricioac has the right schema and we got it wrong in master.

app/models/form_answer_attachment.rb
```diff
@@ -62,10 +62,10 @@ def question_key_correctness
     # possible improvements:
     # - validating all upload questions in this way
     # - smart collecting max-attachments information from the form structure
-    if question_key == "org_chart" && new_record?
```
matthewford: Do we need these changes now that we have a replace working? 

https://github.com/bitzesty/qae/pull/598

Would be good to rebase and test again

mauricioac: @matthewford I don't know... probably we don't need. Will rebase and test again.

mauricioac: @matthewford just rebased and tested, and you're right, works nice with the replace code on master

matthewford:
Ok Closing

mauricioac:
@matthewford just to note: I said that rebasing with master and putting the model validation back it still works. Didn't test only on master without this branch's code. Going to test now

mauricioac:
@matthewford tested locally master branch, and works pretty well, don't think we need this branch

PR #607, opened 2015-09-24
Nominee name data migration task for FormAnswers

dreamfall:
Also removed legacy rake task that said "Remove after use"


rusllonrails:
:+1:

PR #609, opened 2015-09-25
Lead assessors can generate a case status CSV

hammerfunk:
Lead assessors can generate a case status CSV but only for the categories they are a lead in

### DIFF
--- a/app/models/assessor.rb
+++ b/app/models/assessor.rb
@@ -93,13 +93,9 @@ class Assessor < ActiveRecord::Base
   end
 
   def lead_for_any_category?
-    lead = false
-    ["trade", "innovation", "development", "promotion"].each do |cat|
-      if get_role(cat) == "lead"
-        lead = true
-      end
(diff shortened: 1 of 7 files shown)


app/models/assessor.rb
```diff
@@ -92,6 +92,16 @@ def lead_or_assigned?(form_answer)
...
 
+  def lead_for_any_category?
+    lead = false
+    ["trade", "innovation", "development", "promotion"].each do |cat|
+      if get_role(cat) == "lead"
+        lead = true
+      end
+    end
+    return lead
+  end
```
dreamfall:
```ruby
  def lead_for_any_category?
    ["trade", "innovation", "development", "promotion"].any? do |cat|
      get_role(cat) == "lead"
    end
  end
```

hammerfunk: Changed thanks

dreamfall:
:shipit:

PR #610, opened 2015-09-25
Made the Trade C6 questions to be sequential

hammerfunk:
Since we previously removed some questions C 6.1, C 6.2 and C 6.6

We removed C 6.1 and C 6.2 here:
https://github.com/bitzesty/qae/commit/94091f94dd6a1b7c26fbc385ad5e27c4d0740381

And we never added C 6.6 at all:
https://github.com/bitzesty/qae/blob/70382493241af82e0caf4b95c191abae8e5dd8ae/app/forms/qae_2014_forms/international_trade/international_trade_step3.rb#L260


dreamfall:
:+1:

PR #611, opened 2015-09-25
Limited the lead assessors state transition from the state collection

hammerfunk:
Only admins can withdraw or not award an application

### DIFF
--- a/app/models/form_answer_state_transition.rb
+++ b/app/models/form_answer_state_transition.rb
@@ -21,6 +21,12 @@ class FormAnswerStateTransition
   end
 
   def collection
-    form_answer.state_machine.collection(subject)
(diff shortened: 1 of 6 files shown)


app/models/form_answer_state_transition.rb
```diff
@@ -21,6 +21,12 @@ def save
   end
 
   def collection
-    form_answer.state_machine.collection(subject)
+    states = form_answer.state_machine.collection(subject)
+
+    if !subject.kind_of?(Admin) && subject.lead?(form_answer)
```
matthewford: Can normal assessors change the states?

hammerfunk: Nope. Possibly only from submitted to assessment in progress

PR #614, opened 2015-09-28
Conditionals backend cleaning

dreamfall:
Clean up answers on conditional questions upon save if needed in order to avoid displaying them in PDF.


matthewford:
@dreamfall can we squash these?

PR #616, opened 2015-09-28
Form version tweaks

rusllonrails:
Now Form versions would be created by following conditions:

1) Each bump of version would be marked with Person who did update action, like:
```
FormAnswer.last.versions.last.whodunnit
=> "USER:1"
=> "ADMIN:2"
=> "ASSESSOR:13" 
```

2) If it's User's action, then version would be created only once per day


app/controllers/admin/base_controller.rb
```diff
@@ -4,7 +4,7 @@ class Admin::BaseController < ApplicationController
 
   layout "application-admin"
 
-  before_action :authenticate_admin!, :load_award_year_and_settings
+  before_action :authenticate_admin!, :load_award_year_and_settings, :set_paper_trail_whodunnit
```
matthewford: Not needed anymore.

app/models/form_answer.rb
```diff
@@ -5,7 +5,9 @@ class FormAnswer < ActiveRecord::Base
   extend Enumerize
   include FormAnswerStatesHelper
 
-  has_paper_trail if: Proc.new { |t| t.versions.count < 1 || (t.versions.present? && t.versions.last.created_at < Time.zone.now - 1.day) }
+  unless Rails.env.test? # DO NOT NEED TO CREATE VERSIONS IN TEST ENV
```
matthewford: https://github.com/airblade/paper_trail#turning-papertrail-offon Think this is not needed as it is disabled by default for rspec.

app/controllers/admin/base_controller.rb
matthewford: user_for_paper_trail should be enough.

rusllonrails:
@matthewford thanks for review.
Fixed

rusllonrails:
Bravo :tada:

PR #624, opened 2015-09-29
Save on section switch using the sidebar section links

hammerfunk:


app/assets/javascripts/application.js.coffee
```diff
@@ -273,11 +273,12 @@ jQuery ->
   $(document).on "click", ".js-step-link", (e) ->
     e.preventDefault()
     if !$(this).hasClass("step-current")
+      autosave()
+
       current = $(this).attr("data-step")
       if $(this).hasClass "js-next-link"
         if $("body").hasClass("tried-submitting")
           validate()
-        autosave()
```
rusllonrails: @hammerfunk  Just to confirm: we removed autosafe() here because it has no effect - or no need there ?

hammerfunk: It's cos we auto save no matter how we page switch (sidebar or clicking next) so this would be duplication

rusllonrails:
Got it. Good to merge :tada:

PR #626, opened 2015-09-29
previous wins modified to use document attributes

eritiro:
refactoring previous wins to use the same data the user enters in the form.
we stop using previous wins. although it doesn't delete the table.


rusllonrails:
@eritiro  Specs seems failing

app/models/previous_win.rb
```diff
@@ -1,18 +1,10 @@
-class PreviousWin < ActiveRecord::Base
+class PreviousWin
   CATEGORIES = {
     "Innovation" => "innovation",
     "Internation Trade" => "international_trade",
     "Sustainable Development" => "sustainable_development"
   }
 
-  validates :form_answer_id, presence: true
```
rusllonrails: @eritiro not sure, why we are removing validations by category and year there?
If wins no longer stored in table you can use virtual attributes, ActiveModel::Model module and then validate them:

```
include ActiveModel::Model

attr_accessor :category,
                       :year

validates :category, inclusion: {
  in: CATEGORIES.values
}
validates :year, presence: true
```

rusllonrails: If we are not longer support previous wins - why we need PreviousWin model ?

rusllonrails:
@eritiro looks good to merge :+1: 
Just some minor notes above ^

app/models/previous_win.rb
eritiro: It's no longer a ActiveRecord. I just kept it to store some information about the PreviousWinning enums and to stop refactoring at some point (my only concern was the duplicated data structure which is fixed). Of course it would be better to move the list of categories to a more generic class (maybe we already have an enum with the award categories) and the years could be moved to a helper.

PR #627, opened 2015-09-30
Allow collaborators to submit the application form

matthewford:
We have a number of support requests due to the fact that collaborators cannot submit the form, only an admin can.

This reverts the work done to restrict the submission to admins, so that collaborators can submit. We will need to have a wider review of the collaborator role after the submission deadline (today).


rusllonrails:
Good :+1:

PR #629, opened 2015-09-30
Stats

eritiro:
Modified eligibility process to transitionate to "not_eligible".
Created a rake task to take all the non-eligible (calculated in the moment) form_answers and transitionates to "not_eligible"

This PR could be wrong IF:
- There were some rules about what was eligible and what was not. And those rules have changed.
- The logic we use in the model diverge from what the controller actually does.

This could be possible since the model uses 
```
  def eligible?
    eligibility && eligibility.eligible? && (promotion? || (form_basic_eligibility && form_basic_eligibility.eligible?))
  end
```
While the controller uses
```
    builder = FormAnswer::AwardEligibilityBuilder.new(@form_answer)
    @award_eligibility = builder.eligibility
    @basic_eligibility =  builder.basic_eligibility
```

### DIFF
--- a/app/controllers/form_award_eligibilities_controller.rb
+++ b/app/controllers/form_award_eligibilities_controller.rb
@@ -57,6 +57,7 @@ class FormAwardEligibilitiesController < ApplicationController
         if @eligibility.eligible_on_step?(step)
           redirect_to next_wizard_path(form_id: @form_answer.id, skipped: false)
         else
+          @form_answer.state_machine.perform_transition("not_eligible", nil, false) unless @form_answer.not_eligible?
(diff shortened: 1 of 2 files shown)


matthewford:
I guess there are two types of non eligible, we would probably have to do that analysis offline (i.e not in the dashboard) you could answer the basic eligibility as if you are eligible but then when you enter for financials for example, you may find out that actually you are not eligible.

rusllonrails:
@dreamfall also would be great - if you can take a look to it - as you are Eligibility Jedy

rusllonrails:
@eritiro scripts works fine, but then I add new forms (not eligible) - nothing changing in stats dashboard

PR #630, opened 2015-10-01
Stats further fixes

eritiro:
- Added new state for eligibility in progress
- Refactorized transitions between some states
- Refactorized/fixed totals calculations issue


rusllonrails:
Looks fine.
Should be well tested before we apply these changes to servers.
And also would be good to take a look by @dreamfall

PR #636, opened 2015-10-05
Removed the word count paste JS as it was hijacking the natural paste

hammerfunk:
I previously writ this code as the word count wasn’t working for paste,
but just tested this without this code and it’s now totally fine even
on old IE


rusllonrails:
Cool :+1:

PR #645, opened 2015-10-08
display country name rather than ISO code in CSV reports

dreamfall:
add missing immediate parent company name to the Entries report

https://trello.com/c/r5BxPByw/100-entries-report-missing-immediate-parent-name-and-country-should-be-full-name


matthewford:
@dreamfall needs a rebase

PR #647, opened 2015-10-09
display country name rather than ISO code in CSV reports

dreamfall:
add missing immediat parent company name to the Entries report


rusllonrails:
Great :+1:

PR #648, opened 2015-10-09
display country name not ISO code in OrganisationWithUltimateControlCountry column

dreamfall:
https://trello.com/c/r5BxPByw/100-entries-report-missing-immediate-parent-name-and-country-should-be-full-name


rusllonrails:
Good :+1:

PR #654, opened 2015-10-21
Primary assessor permissions

hammerfunk:
Allow primary assessors:
- see the moderated appraisal after moderated appraisal submission
- edit the case summary

### DIFF
--- a/app/helpers/partials_visibility_helper.rb
+++ b/app/helpers/partials_visibility_helper.rb
@@ -5,7 +5,7 @@ module PartialsVisibilityHelper
   end
 
   def show_section_appraisal_moderated?
-    current_subject.lead?(@form_answer)
+    policy(@form_answer).show_section_appraisal_moderated?
   end
 
   def show_winners_section?
--- a/app/models/assessor_assignment.rb
+++ b/app/models/assessor_assignment.rb
@@ -92,6 +92,7 @@ class AssessorAssignment < ActiveRecord::Base
   def owner_or_administrative?(subject)
     subject.is_a?(Admin) ||
       subject.try(:lead?, form_answer) ||
+      (case_summary? && subject.try(:primary?, form_answer)) ||
       (!moderated? && !case_summary? && assessor_id == subject.id)
   end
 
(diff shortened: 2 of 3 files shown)


hammerfunk:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Qky4rAbR&#x2F;217-primary-assessor-should-have-the-case-summary-view-as-well-as-they-need-to-be-able-to-edit-it-etc-for-the-panels-going-forward"> Primary Assessor should have the case summary view as well as they need to be able to edit it etc for the Panels going forward</a>

app/helpers/partials_visibility_helper.rb
```diff
@@ -5,7 +5,9 @@ def show_section_case_summary?
   end
 
   def show_section_appraisal_moderated?
-    current_subject.lead?(@form_answer)
+    return true if current_subject.lead?(@form_answer)
+    return false unless moderated_assessment.submitted?
+    return true if current_subject.primary?(@form_answer)
```
eritiro: I would replace the method content with the line below, because I think it's more clear. but maybe It's s just me. In that case ignore this comment. 
current_subject.lead?(@form_answer) || (moderated_assessment.submitted? && current_subject.primary?(@form_answer))

eritiro: I think it returns nil if no one of the three conditions are met.

matthewford: why is this in a helper would it not be better of in the permissions?

hammerfunk: I have no idea. I'll make that change Emiliano :)

hammerfunk: Ok that code doesn't work so well cos it can't tell if what the moderated assessment is
```
undefined local variable or method `moderated_assessment' for #<AssessorAssignment:0x007feb9fc083b0>
```
Also I included the admin and changed the rest to sound like it was before except that moderated assessment code
```
    subject.is_a?(Admin) ||
      subject.try(:lead?, form_answer) ||
      (moderated_assessment.submitted? && subject.try(:primary?, form_answer))
```

eritiro:
Great job!

PR #658, opened 2015-10-29
Added async worker for custom_email_form

eritiro:
- fixed typo in custom_emails params
- ignored errors during custom email batch
- added raven exception to track those errors.
- modified raven configuration to explicitly set environments

### DIFF
--- a/app/workers/custom_email_worker.rb
+++ b/app/workers/custom_email_worker.rb
@@ -1,0 +1,10 @@
+class CustomEmailWorker
+  include Shoryuken::Worker
+
+  shoryuken_options queue: "#{Rails.env}_custom_emails", auto_delete: true
+
+  def perform(sqs_msg, request)
+    puts "processing email #{request}"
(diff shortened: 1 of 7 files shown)


app/workers/custom_email_worker.rb
```diff
@@ -0,0 +1,10 @@
+class CustomEmailWorker
+  include Shoryuken::Worker
+
+  shoryuken_options queue: "#{Rails.env}_custom_emails", auto_delete: true
```
rusllonrails: @eritiro 
Looks fine, but do we really need extra queue for Custom Emails?

We have 2 queues:
1) "#{Rails.env}_mailers" - for all mailers
2) "#{Rails.env}_default" - for jobs

What if we just put it into:
```
shoryuken_options queue: "#{Rails.env}_default", auto_delete: true
```

rusllonrails:
Looks good :+1:

PR #671, opened 2015-11-13
[Audit Certificate] fixes by client notes

rusllonrails:
Related to [Trello Story](https://trello.com/c/zSuswff6/228-for-the-audit-certificate-template-change-the-financials-table-to-be-the-same-as-the-financial-summary-in-the-admin-view)

1. The "percentage growth" grids and the "overall growth" grids are not required on the audit templates
2. The audit templates are not required for applicants who are 'Not recommeded'.
   They are only for the shortlisted entrants i.e. those who are 'Recommended' or 'Reserved'

### DIFF
--- a/app/policies/form_answer_policy.rb
+++ b/app/policies/form_answer_policy.rb
@@ -68,6 +68,6 @@ class FormAnswerPolicy < ApplicationPolicy
   end
 
   def can_download_initial_pdf?
-    admin?
+    admin? && record.shortlisted?
   end
 end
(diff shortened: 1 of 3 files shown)


app/policies/form_answer_policy.rb
```diff
@@ -68,6 +68,6 @@ def has_access_to_post_shortlisting_docs?
   end
 
   def can_download_initial_pdf?
```
matthewford: shall we also rename this policy to be more reflective of what it does then? can download audit certificate?

rusllonrails: agree, will fix

PR #675, opened 2015-11-17
[International Form] Financial Summary and Audit Certificate tweaks

rusllonrails:
[Trello Story](https://trello.com/c/3VPDVlC4/231-for-international-trade-always-display-the-two-optional-fields-in-the-financial-summary-and-audit-certificate-even-if-blank)

For International Trade, always display the two optional fields
in the financial summary and audit certificate, even if blank.

### DIFF
--- a/app/form_pointers/form_financial_pointer.rb
+++ b/app/form_pointers/form_financial_pointer.rb
@@ -26,6 +26,10 @@ class FormFinancialPointer
     :trade,
     :promotion
   ]
+  TRADE_OPTIONAL_OPS = [
+    :total_imported_cost,
+    :overseas_yearly_percentage
+  ]
 
   def initialize(form_answer, options={})
     @form_answer = form_answer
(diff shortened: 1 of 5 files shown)


app/form_pointers/form_financial_pointer.rb
```diff
@@ -255,6 +265,46 @@ def average_growth_for(form_answer, year)
...
+            name: "#{option}_#{i + 1}of#{years_length}",
+            value: ""
+          }
+        end
+
+        option_data_hash[option] = option_data_values
+
+        if option == :total_imported_cost &&
+           fetched.detect { |el| el.has_key?(:overseas_yearly_percentage) }.present?
+          fetched.insert(-2, option_data_hash)
```
matthewford: why `-2`?

rusllonrails: @matthewford so because questions:
Total Imported Cost should go before Overseas Yearly Percentage.
These question are not depending from each other.
For example:
   form can have overseas_yearly_percentage but have not total_imported_cost
   In this case we need to populate total_imported_cost with empty values.
   If we will use << array method or .insert(-1, option_data_hash)
   then Total Imported Cost  would go after Overseas Yearly Percentage.

matthewford: ok

PR #676, opened 2015-11-18
[Form Dashboard] Investigate why user cannot see their past applications

rusllonrails:
[Trello Story](https://trello.com/c/QZa0iOQ6/233-investigate-why-alister-dell-metricell-com-cannot-see-their-past-applications)

So applications can have status of shortlisted states ('reserved' and 'recommended')
before system settings transition to shortlisted mode.
System settings are shortlisted once Admin will schedule Shortlisted Mailer.
System were displaying only applications in "assessment_in_progress submitted not_submitted application_in_progress"
states, because system settings are not in after_shortlisting_stage.
So that some applications with states ('reserved' and 'recommended') were missing
in Dashboard.

### DIFF
--- a/app/models/form_answer_transition.rb
+++ b/app/models/form_answer_transition.rb
@@ -1,6 +1,15 @@
 class FormAnswerTransition < ActiveRecord::Base
   include Statesman::Adapters::ActiveRecordTransition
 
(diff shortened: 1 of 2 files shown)


dreamfall:
:+1:

PR #677, opened 2015-11-18
[Admin | Entries Report] add 'Overall Status' column

rusllonrails:
[Trello Story](https://trello.com/c/wyxirrHS/234-add-in-a-new-column-in-entries-report-csv)

An additional column in the 'Entries' report that displays the grade
in the white "Status" section in the Assessment site.
This is so that we can compare the grade in the new column with
the one in the existing 'CaseSummaryOverallGrade' to see
if there any differences in the two grades.


dreamfall:
:shipit:

PR #678, opened 2015-11-18
[Form PDF and Read Only View] Make the font colour darker so its easier to read

rusllonrails:
[Trello Story](https://trello.com/c/9WyzDrar/232-make-the-font-colour-on-the-pdf-for-test-and-the-read-only-form-darker-so-its-easier-to-read)


eritiro:
looks good!

PR #696, opened 2015-12-07
[Admin] ability to remove Audit Certificate

rusllonrails:
[Trello Story](https://trello.com/c/enSgNSte/248-qae-admins-can-delete-the-auditors-certificate-in-case-the-user-has-uploaded-it-in-error)

Some users have uploaded the blank form, and the admins need to be deleted.
And the status will need to go back to incomplete so they can reupload it.
Also need to updated the cases status report so it says the certificate has not been uploaded.

### DIFF
--- a/app/assets/javascripts/admin/form_answers.js.coffee
+++ b/app/assets/javascripts/admin/form_answers.js.coffee
@@ -13,6 +13,7 @@ ready = ->
   handleWinnersForm()
   handleReviewAuditCertificate()
   handleReviewCorpResponsibility()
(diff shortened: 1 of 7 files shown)


rusllonrails:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;enSgNSte&#x2F;248-qae-admins-can-delete-the-auditors-certificate-in-case-the-user-has-uploaded-it-in-error"> QAE: Admins can delete the auditors certificate in case the user has uploaded it in error.</a>

PR #698, opened 2015-12-07
[Admin / Assessor] simple_format helper issue

rusllonrails:
Related to [Trello Story](https://trello.com/c/DnykSDMv/246-qae-issue-with-qa0590-16s-admins-should-be-able-to-change-verdict-on-moderated-forms)

So we use simple_format helper for displaying linebreaks in assessor / admin
comments/ drafts and other 'textarea' content blocks.
http://apidock.com/rails/v4.2.1/ActionView/Helpers/TextHelper/simple_format

Problem is that simple_format always adds wrapper tags (p - by default)
It causes issue on JS code after clicking on 'Save'. (Attached screenshot below)
And there are no way to skip it as if you pass empty 'wrapper_tag' option like:

```
simple_format text, {}, wrapper_tag: ""
```
it will add brackets without any tag:
```
<>
```

So that's we overrided source code of simple_format and now
will use own qae_simple_format helper.

### DIFF
--- a/app/assets/javascripts/admin/form_answers.js.coffee
+++ b/app/assets/javascripts/admin/form_answers.js.coffee
@@ -165,7 +165,7 @@ ready = ->
     formGroup.removeClass("form-edit")
 
     if area.val().length
-      formGroup.find(".form-value p").text(area.val())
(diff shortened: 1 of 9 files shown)


eritiro:
looks ok!

PR #711, opened 2016-01-12
change CaseSummaryStatus column logic in Cases Status report

dreamfall:


dreamfall:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;qi6si3ug&#x2F;507-qae-queen-s-awards-cases-status-report"> QAE: Queen&#39;s Awards Cases Status Report</a>

app/models/reports/form_answer.rb
```diff
@@ -191,12 +191,10 @@ def qao_agreed_press_note
   end
 
   def case_summary_status
-    lead_submitted = @case_summary.try(:submitted?)
-
-    if !lead_submitted
+    if @case_summary.try(:submitted?) && @case_summary.locked?
```
rusllonrails: @dreamfall I guess it's enaught to just use:
```
if @case_summary.try(:submitted?)
   ...
end
```
because each time then application is submitted or resubmitted -> locked_at would be set with current DateTime.
https://github.com/bitzesty/qae/blob/master/app/services/assessment_submission_service.rb#L44

rusllonrails:
Looks good :+1:

PR #719, opened 2016-01-20
better healthcheck message

matthewford:
catch missing migrations and change success message to something stable


rusllonrails:
:+1:

PR #722, opened 2016-01-21
Merge unsuccessful emails

eritiro:
all_unsuccessful and  unsuccessful email notifications became one unsuccessful notification. 

all_unsuccessful logic was kept (using different emails for ep/non-ep and without excluding forms without audit_certificate) but unsuccessful kind was preserved (there is a new task that migrates all_unsuccessful to unsuccessful)


rusllonrails:
Looks ok.

PR #740, opened 2016-02-18
[Admin Settings] Populate of new buckingham palace's deadlines

rusllonrails:
Rake script adds new missing deadlines to current settings:

* media information
* confirm press book notes
* attendees invite

Run:

```
bundle exec rake db:populate_missing_buckingham_palace_deadlines
```

[Trello Story](https://trello.com/c/TVZVlcMT/270-qae-support-change-unsuccessful-ep-email-copy-for-as-they-do-not-get-feedback)

### DIFF
--- a/lib/tasks/temprary/populate_missing_buckingham_palace_deadlines.rake
+++ b/lib/tasks/temprary/populate_missing_buckingham_palace_deadlines.rake
@@ -1,0 +1,16 @@
+namespace :db do
+  desc "Populate of new buckingham palace's deadlines (media information,  confirm press book notes,  attendees invite) on current Settings"
+  task populate_missing_buckingham_palace_deadlines: :environment do
+    [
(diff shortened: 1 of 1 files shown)


macool:
@rusllonrails there's also `Setting#create_deadlines`

https://github.com/bitzesty/qae/blob/master/app/models/settings.rb#L68

who creates missing deadlines

https://github.com/bitzesty/qae/blob/master/app/models/deadline.rb#L9

macool:
rake task would be something like `Setting.find_each {|s| s.send :create_deadlines }`

rusllonrails:
@macool thanks, 
yeah I know it, but tbh I doesn't trust to find_or_create!, can't explain why =)
maybe because I got some issues with find_or_create! on localhost.
So that I decided to add rake script which adds only missing. =)

PR #749, opened 2016-02-22
copytext / ui updates

macool:
includes points 5, 6, 7 & 8 from provided document

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160222153821) do
+ActiveRecord::Schema.define(version: 20160222175452) do
 
   # These are extensions that must be enabled in order to support this database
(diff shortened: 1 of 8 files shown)


macool:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;YGQ6smF1&#x2F;643-qae-support-successful-business-winners-press-book-entry-screens"> QAE Support: Successful Business Winners Press Book Entry Screens</a>

db/schema.rb
```diff
@@ -312,7 +312,7 @@
     t.boolean  "approved",         default: false
     t.datetime "created_at",                       null: false
     t.datetime "updated_at",                       null: false
-    t.string   "first_name"
+    t.text     "first_name"
```
matthewford: Where is this from? Is this needed?

matthewford: A string will be fine, we can revert this change

matthewford: perhaps rename it to name

PR #755, opened 2016-02-24
[Application] Unsuccessful state corrects

rusllonrails:
[Trello Story](https://trello.com/c/TVZVlcMT/270-qae-support-change-unsuccessful-ep-email-copy-for-as-they-do-not-get-feedback)


rusllonrails:
Not needed

PR #759, opened 2016-02-24
Exclude reserved applications from unsuccessful email list

eritiro:
According to the last comment from Kerry:

> Just to clarify again for QAEP, International Trade, Innovation and Sustainable Development the unsuccessful emails should go to those where the "STATUS" in the white section on the left hand side says "Not Recommended" or "Not Awarded".
> Where the STATUS is "Withdrawn" they do not get anything.

### DIFF
--- a/app/models/form_answer.rb
+++ b/app/models/form_answer.rb
@@ -110,7 +110,7 @@ class FormAnswer < ActiveRecord::Base
     scope :shortlisted, -> { where(state: %w(reserved recommended)) }
     scope :not_shortlisted, -> { where(state: "not_recommended") }
     scope :winners, -> { where(state: "awarded") }
-    scope :unsuccessful, -> { where(state: %w(not_recommended not_awarded reserved)) }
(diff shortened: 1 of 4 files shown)


app/models/form_answer.rb
```diff
@@ -110,7 +110,7 @@ def secondary
     scope :shortlisted, -> { where(state: %w(reserved recommended)) }
     scope :not_shortlisted, -> { where(state: "not_recommended") }
     scope :winners, -> { where(state: "awarded") }
-    scope :unsuccessful, -> { where(state: %w(not_recommended not_awarded reserved)) }
+    scope :unsuccessful, -> { where(state: %w(not_recommended not_awarded)) }
```
matthewford: See comments on cards - I think this should be, where ! awarded or withdrawn

eritiro: Ok. sec.

eritiro:
@matthewford done!

PR #767, opened 2016-03-01
Collaborators receive email notifications

eritiro:
- the system sends a notification to unsuccessful EP  and non-EP applications users and collaborators
- the system sends a notification to winners to check their press release to EP  and non-EP applications users and collaborators.
- we no longer send notifications to the user when an admin approves a press release


eritiro:
https://trello.com/c/XuEdoOAG/290-qae-support-confirm-which-emails-are-sent-to-which-users

rusllonrails:
Looks good :+1: , but need corrects, explained in slack

PR #773, opened 2016-03-03
[DO NOT MERGE] Vigilion integration

eritiro:
- schema and data migrations
- updated Scan references to make use of the vigilion scan_results columns
- removed unused clases and fixed specs


matthewford:
@macool can you make this against the server-maintenance branch instead so we can close this one?

matthewford:
closing as it should be against the CF branch @macool

macool:
@matthewford :indeed: will rebase

PR #782, opened 2016-03-07
do not allow unlocking press summary if on winners stage

macool:
https://trello.com/c/GSA6Lb8h/697-qae-support-once-winners-email-is-sent-out-assessors-should-not-be-able-to-unlock-the-pressbook-section

### DIFF
--- a/app/policies/press_summary_policy.rb
+++ b/app/policies/press_summary_policy.rb
@@ -18,7 +18,7 @@ class PressSummaryPolicy < ApplicationPolicy
   end
 
   def unlock?
-    record.submitted? && subject.lead?(form_answer)
+    record.submitted? && subject.lead?(form_answer) && !Settings.winners_stage?
   end
 
   private


macool:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GSA6Lb8h&#x2F;697-qae-support-once-winners-email-is-sent-out-assessors-should-not-be-able-to-unlock-the-pressbook-section"> QAE Support: Once Winners email is sent out Assessors should not be able to unlock the pressbook section</a>

app/views/admin/press_summaries/_section.html.slim
```diff
@@ -40,7 +40,7 @@
         p.p-empty Applicant hasn't confirmed the Press Book Notes
 
 - if press_summary.submitted?
-  - if policy(press_summary).unlock?
+  - if policy(press_summary).unlock? && !Settings.winners_stage?
```
matthewford: Can we not just include this in the existing policy rather than adding an additional check?

matthewford: @macool See: https://github.com/bitzesty/qae/pull/780/files

app/controllers/concerns/press_summary_mixin.rb
```diff
@@ -4,6 +4,7 @@ module PressSummaryMixin
   def self.included(base)
     base.before_action :load_form_answer
     base.before_action :load_press_summary, except: :create
+    base.before_action :verify_not_in_winners_stage, only: :unlock
```
matthewford: dont know if you need this - should be covered by the policy

macool: I'll make sure now

macool: you're right

PR #788, opened 2016-03-09
run under convox/docker

macool:
```
convox start
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,37 @@
+FROM gliderlabs/alpine:3.3
+
+ENV BUILD_PACKAGES="curl-dev ruby-dev build-base git" \
+    DEV_PACKAGES="zlib-dev libxml2-dev libxslt-dev libffi-dev tzdata yaml-dev libpq curl qt5-qtbase qt5-qtbase-dev qt5-qtwebkit-dev postgresql-client postgresql-dev" \
+    RUBY_PACKAGES="ruby ruby-irb ruby-io-console ruby-bigdecimal ruby-json yaml nodejs" \
+    RAILS_VERSION="4.2.5.1" \
+    PATH="$PATH:/usr/lib/qt5/bin/"
+
+RUN \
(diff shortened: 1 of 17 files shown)


macool:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gYPar1oC&#x2F;650-qae-support-server-maintenance"> QAE Support: Server Maintenance</a>

Dockerfile
```diff
@@ -0,0 +1,32 @@
+FROM phusion/baseimage:latest
```
matthewford: Image is too large, can we use the gliderlabs/alpine image

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+    - 3000:3000
+  volumes:
+    - .:/app
+  links:
+    - db
+    - redis
+worker:
+  <<: *web
+  ports: []
+  command: bundle exec sidekiq -L ./log/worker.log
```
matthewford: should log to stdout

matthewford:
Can we change the docker_compose file to `docker_compose.yml.local` which you can specify on start. 

This is because this is the file convox uses on deployment.

macool:
Do not merge please. I'm still failing with

`web    | /usr/lib/ruby/gems/2.2.0/gems/bundler-1.11.2/lib/bundler/runtime.rb:80:in `rescue in block (2 levels) in require': There was an error while trying to load the gem 'skylight'. (Bundler::GemRequireError)`

matthewford:
Remove skylight, it uses rust so could be missing a dependency.

sent from my phone.

macool:
@matthewford I simplified circle file by adding `sql` structure dump

config/application.rb
```diff
@@ -33,6 +33,7 @@ class Application < Rails::Application
     end
 
     config.cache_store = :memory_store
+    config.active_record.schema_format = :sql
```
matthewford: @macool don't we need to update the schema then?

matthewford: Ignore me, I didn;t see the file as github hid it.

PR #790, opened 2016-03-10
[Press Summary] bunch of corrects and improvements

rusllonrails:
[Trello Story](https://trello.com/c/P1xMEUpm/289-qae-support-successful-business-winners-press-book-entry-screens)

1. Top line i.e. ‘Congratulations on winning a Queen's Award for Enterprise’ – to be bolded (see screen shot 1)

2. Also in the first paragraph underneath that the words: "strict embargo" and also "until 00.01 on Thursday 21 April 2016.
   You must not make any announcements about your Award,
   either to employees or outside your organisation, before this date." are to be bolded

3. With the ‘Head of organisation’ details if the person has a personal “Honour”
   it is not being picked up and displayed in the users view (see screen shot 2 & example test case QA0030/16T)

4. I mentioned before on this card (8 March) that we need a "Save" button
   (see attached screen shot) as otherwise the applicant needs to enter their data all in one go.
  Can we perhaps have it next to the "Submit" button (see screen shot 3)

5. Once an applicant has submitted their Press entry details for a particular case,
   although the button turns to say “COMPLETE”(see screen shot 5) we have found that
   they can still do further changes /edits and re-submit. This is not to be allowed.
   Once they have submitted for that particular entry it should lock, otherwise this will casue problems at our end.

6. In the instances whereby the applicant does not respond to the exercise by the deadline,
   then the system, although allowing them to view the details should not alow them to respond, i.e. the system locks to them.

7. For those who do not respond by the “Press Book Entry confirmation due” deadline,
   QA admins need to have the 3 boxes for the Press contact details (all be it blank)
   in the assessment site of the entry so that we can manually populate it and then submit. (see scren shot 4 but as I say the boxes would be blank)

8. Users view should not display any QAEP successful or unsuccesssful entries
   (see screen shot 5) that the account holder may have

### DIFF
--- a/app/policies/press_summary_policy.rb
+++ b/app/policies/press_summary_policy.rb
@@ -21,6 +21,29 @@ class PressSummaryPolicy < ApplicationPolicy
     record.submitted? && subject.lead?(form_answer) && !Settings.winners_stage?
   end
 
+  def can_see_contact_details?
+    record.reviewed_by_user? || deadline_passed?
+  end
+
+  def can_update_contact_details?
+    (
+      !deadline_passed? &&
+      update?
+    ) ||
+    (
+      deadline_passed? &&
+      admin?
+    )
+  end
+
+  def deadline_passed?
+    Settings.current
+            .deadlines
+            .where(kind: "buckingham_palace_confirm_press_book_notes")
+            .first
+            .passed?
+  end
+
   private
 
   def form_answer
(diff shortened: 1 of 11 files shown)


app/policies/press_summary_policy.rb
```diff
@@ -21,6 +21,29 @@ def unlock?
...
+  end
+
+  def can_update_contact_details?
+    (
+      !deadline_passed? &&
+      update?
+    ) ||
+    (
+      deadline_passed? &&
+      admin?
```
matthewford: This can probably be just `(deadline_passed? && update?) || admin?`

rusllonrails: @matthewford  agree) Will fix

PR #806, opened 2016-03-14
add winner emblems & manual for 2016 year winners

macool:
@matthewford I made a new folder to place 2016 assets and we now have different templates for each award year. however the controller will render only current award year as we'd still need to implement some filters in controllers so we query only current year's awards. eg https://github.com/bitzesty/qae/blob/28ed8a7cce013453841fcd735662cdda03642e86/app/controllers/content_only_controller.rb#L59


macool:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;tMEGFgto&#x2F;701-qae-support-update-winners-sections-with-new-emblems-manual"> QAE Support: Update winners sections with new emblems &amp; manual</a>

PR #812, opened 2016-03-16
use sidekiq-cron instead of whenever

macool:


macool:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gYPar1oC&#x2F;650-qae-support-server-maintenance"> QAE Support: Server Maintenance</a>

app/workers/scheduled/performance_platform_service_worker.rb
```diff
@@ -0,0 +1,9 @@
+module Scheduled
+  class PerformancePlatformServiceWorker < BaseWorker
+    def perform
+      ::CronJob.run("performance") do
```
matthewford: Can remove this class - it was used for locking - it is no longer needed

config/initializers/sidekiq-cron.rb
```diff
@@ -0,0 +1,18 @@
...
+
+production_schedule = {
+  "performance_platform_service" => {
+    "cron" => "0 0 * * 0",
+    "class" => "Scheduled::PerformancePlatformServiceWorker"
+  }
+}
+
+Sidekiq::Cron::Job.load_from_hash(default_schedule)
+if Rails.env.production?
```
matthewford: Probably want to use an ENV Variable as all apps will run in production env

Gemfile.lock
```diff
@@ -413,6 +412,8 @@ GEM
     redis-activesupport (4.0.0)
       activesupport (~> 4)
       redis-store (~> 1.1.0)
+    redis-namespace (1.5.2)
```
matthewford: :( do we need this?

macool: @matthewford I upgraded sidekiq but it's a dependency for `sidekiq-cron` gem

PR #831, opened 2016-04-08
safely paste from word 

eritiro:
javascript that intercept the paste event on textareas and removes word-specific characters in order to safely import text from word

### DIFF
--- a/app/assets/javascripts/clean-paste.js
+++ b/app/assets/javascripts/clean-paste.js
@@ -1,0 +1,13 @@
+$(function(){
+  $("textarea").bind('paste',function(e){
+    var rte = $(this);
+    setTimeout(function(){
+        var text = rte.val();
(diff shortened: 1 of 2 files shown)


eritiro:
Since this is a kind of manually "replace", It would be nice to test it with a client example.

eritiro:
https://trello.com/c/a83j5Nhw/766-qae2015-strip-ms-word-specific-characters

rusllonrails:
Looks reasonably well :+1:

app/assets/javascripts/clean-paste.js
```diff
@@ -0,0 +1,13 @@
+$(function(){
+  $("textarea").bind('paste',function(e){
+    var rte = $(this);
+    setTimeout(function(){
+        var text = rte.val();
+        text = text.replace(/(.*(?:endif-->))|([ ]?<[^>]*>[ ]?)|(&nbsp;)|([^}]*})/g,'');
```
matthewford: where is this list from. I think there are more of these symbols.

PR #894, opened 2016-04-28
[WIP] Temporal workaround to bypass validation

theharq:
#### What this PR does:

Bypass the drops validation on the International Trade form when the question C 2.1 is answered "yes"

xxx

#### Notes

This is a workaround if this change needs to enter as soon as possible, but it is not the best way to solve this problem.

Merge only if you are ok with this workaround.


matthewford:
is this check only on the frontend?

theharq:
As far I could see in the code, this particular validation was made in the client only.. but I will have to check where and how this information is persisted.. I suggest wait for @rusllonrails  feedback on this issue.

rusllonrails:
Well done :+1:

PR #918, opened 2016-05-03
Update formatted trigger time decorator

theharq:
#### What this PR does:

- More updates to time format.
From `%H:%M` to `%-l:%M%P`    Ex: `14:08` becomes `2:08pm`
- Move helpers methods to a Decorator class.
- Swap strategy for database_cleaner

#### Notes

>The documentation says:

>For the SQL libraries the fastest option will be to use :transaction as transactions are simply rolled back. If you can use this strategy you should.

We can set the truncation strategy only for js tests since capybara will run in another thread and will not share the connection to rollback a transaction.

### DIFF
--- a/app/decorators/application_decorator.rb
+++ b/app/decorators/application_decorator.rb
@@ -6,7 +6,7 @@ class ApplicationDecorator < Draper::Decorator
   end
 
   def created_at_timestamp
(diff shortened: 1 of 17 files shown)


rusllonrails:
:+1: Well done

PR #919, opened 2016-05-04
[WIP] Rspec updates

theharq:
#### What this PR does:

- Adds the missing `spec_helper` and updates the `rails_helper` files.
- Remove `Capybara` debugging line and replace the TODO comment with a pending spec
- Speed up "filters by sub options" spec from 39.62 to 4.5 seconds
- Speed up "filters by status" spec from 13.72 to 4.39 seconds

#### Notes

Because the addition of the `spec_helper` which have this config line: `mocks.verify_partial_doubles = true` we could show that we have 2 false positives, when stubbing methods that did not exists in the real classes.

We also have the possibility now to check for slow specs un-commenting the line `config.profile_examples = 10` in the `spec_helper` file.


spec/features/users/form_answers/past_applications_spec.rb
```diff
@@ -65,10 +65,8 @@
         visit dashboard_path
       end
 
-      it "should display past successful applications" do
-        save_and_open_page
-
-        # TODO: need to finish this spec
+      xit "should display past successful applications" do
+        # save_and_open_page
```
rusllonrails: yeah - it's mine "super usefull" spec =)

rusllonrails:
Good Job :+1:

PR #938, opened 2016-06-09
fix default answer picker

dreamfall:


dreamfall:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;BtYq55Tq&#x2F;964-qae16imp-question-c1-answer-changing-itself"> [QAE16IMP] Question C1 answer changing itself</a>

PR #941, opened 2016-06-09
Correctly load test env variables

theharq:
#### What this PR does:

- Remove foreman from the Gemfile
- Update setup script to check if foreman is installed in the system
- Delete spec file, no longer needed

#### Notes

Foreman should not be present in the Gemfile:

> “Ruby users should take care not to install foreman in their project's Gemfile”
https://github.com/ddollar/foreman

I moved `rspec-rails` gem to the dev and test bundler groups

> Add rspec-rails to both the :development and :test groups in the Gemfile
https://github.com/rspec/rspec-rails


matthewford:
@theharq actually, thinking about it we can remove forman and the profile all together

PR #944, opened 2016-06-10
Convox fixes

theharq:
#### What this PR does:

Convox fixes:

- Do not install gems for development mode
- Add env variables that may cause issues if they are not present, like when building from CircleCI, during the asset precompilation

#### Notes

I've added a `convox-start.yml` file to be used with `convox start -f convox-start.yml`, useful if we want to run the app in production mode.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,5 +28,6 @@ COPY . /app
 
 ENV RAILS_ENV=production
 ENV ONLY_ASSETS=true
+ENV DATABASE_URL postgresql://localhost/dummy_url
 
 RUN bundle exec rake assets:precompile --trace
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,7 +12,6 @@ RUN apk add --update --no-cache \
     git \
     postgresql-client \
     postgresql-dev
-RUN bundle config build.nokogiri --use-system-libraries
 
 EXPOSE 3000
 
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -12,6 +12,7 @@ RUN apk add --update --no-cache \
     git \
     postgresql-client \
     postgresql-dev
+RUN bundle config build.nokogiri --use-system-libraries
```
matthewford: I had to remove this for nokogiri to build on alpine.

https://github.com/gliderlabs/docker-alpine/issues/53#issuecomment-224548821

matthewford:
@theharq shall we remove then `docker-compose.yml.local` and update the readme for the convox start yaml file?

theharq:
I think is ok to preserve the `docker-compose.yml.local` which is meant to be used when developing the app with docker, but I don't know if any dev of the team will do that.. is that is desired.. cool, if not, we could delete the file because YAGNI

matthewford:
not sure I see much difference between having a convox start file and a compose local file..

theharq:
well.. the difference is the mode how it starts.. there are 2 little differences... one mounts the app folder and runs the app in development mode.. the other one is to simulate production.. but I think we can leave only one.. what file do you prefer to stay.. `docker-compose.yml.local` or `convox-start.yml` I'm ok whatever file is 👍

PR #946, opened 2016-06-13
Add in QAE module for environment variable detection 

matthewford:


config/environments/staging.rb
```diff
@@ -1,95 +0,0 @@
-Rails.application.configure do
```
rusllonrails: @matthewford why we are going to remove config/environments/staging.rb ?

rusllonrails:
Looks good :+1: 
One small question above ^

config/environments/staging.rb
matthewford: @rusllonrails as per 12factor, all env run in production with ENV variables for the different settings.

rusllonrails: Ok. fine

PR #949, opened 2016-06-16
Updates for convox

theharq:
#### What this PR does:

- Remove dependency of the `ONLY_ASSETS` env variable used to disable blocks of code that are not supposed to run durint the boot process when precompiling the assets.
- Remove the rails `staging` environment
- Remove unnecessary code from initializers
- Update `database.yml` file to read from the `DATABASE_URL` as the convention dictates.
- Add required AWS keys to the `.env` sample file

#### Notes

I was looking into the source code of the `sprocket-rails` and I noticed that in all the rake task that this gems add, like `assets:precompile` the `assets:environment` task is called first and default this loads the Rails environment. 

Since we are building the docker image, we should not have code in the initializers that attemps to connect to the db in order to avoid issues.


Gemfile
```diff
@@ -144,3 +121,26 @@ group :development, :test do
   gem 'dotenv-rails'
   gem 'rspec-rails', '~> 3.4'
 end
+
+group :production do
+  # Error reporting
+  gem 'sentry-raven', github: 'getsentry/raven-ruby'
```
matthewford: we can now use the released version

config/initializers/paper_trail.rb
```diff
@@ -1,5 +1,3 @@
-unless ENV["ONLY_ASSETS"]
```
matthewford: papertrail update may mean we don't need this anymore, but this was causing a db connection on asset compilation

matthewford:
Looking good.

theharq:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;LUfiXGdw&#x2F;970-qae16imp-update-docker-image-to-not-rely-on-environment-variables-for-asset-compilation"> [QAE16IMP] Update docker image to not rely on environment variables for asset compilation</a>

PR #983, opened 2016-07-27
add bootscale gem

dreamfall:
Rails server starts ~2s faster with bootscale. 

https://github.com/byroot/bootscale


rusllonrails:
@matthewford what do you think about bootscale gem?

rusllonrails:
:rocket:

PR #1014, opened 2016-08-09
add offline library to track user connection to the server

dreamfall:


dreamfall:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;FUHBJtkj&#x2F;1074-qae16imp-lost-connection-error-message"> [QAE16IMP] Lost connection error message</a>

matthewford:
We have websockets build it for presence checking, we should use that

PR #1035, opened 2016-08-15
[Appraisal Form] remove strengths keys from form

rusllonrails:


rusllonrails:
Specs failing - will review and prepare another PR

PR #1059, opened 2016-08-26
Modify year-end dates only for empty values

theharq:
#### What this PR does:

Modify the year end dates values only for those fields which are empty

Before of this change the script modified the previous value entered by the user.

years


theharq:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zHboyypr&#x2F;1170-qae16imp-question-c2-2-innovation-financial-year-end-dates"> [QAE16IMP] Question C2.2 - Innovation - financial year end dates</a>

PR #1075, opened 2016-08-31
[Hotfix] Move out summary and feedback sections to be visible by default

theharq:
Taking case summary out of assessment section, and making it as well as the feedback section visible at all times

case


theharq:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;rTwMUSvx&#x2F;1185-qae16imp-feedback-and-case-summary"> [QAE16IMP] Feedback and Case Summary</a>

PR #1076, opened 2016-08-31
Fixing date format on year dates questions

mauricioac:
The en dates per year questions don't validate if user is typing "05". So on PDFs the dates end up being not "padded".

This PR addresses that by padding the day and month of these dates.

### DIFF
--- a/app/form_pointers/financial_table.rb
+++ b/app/form_pointers/financial_table.rb
@@ -41,21 +41,13 @@ module FinancialTable
       corrected_result << if entry.size == 6 && (index + 1) != res.size
         year = last_year - (res.size - (index + 1))
         entry += year.to_s
-        fix_day_month_format(entry)
+        entry
       else
-        fix_day_month_format(entry)
+        entry
       end
     end
 
-    corrected_result
-  end
-
-  def fix_day_month_format(date)
-    parts = date.split("/")
-    day = parts[0].rjust(2, "0")
-    month = parts[1].rjust(2, "0")
-
-    "#{day}/#{month}/#{parts[2]}"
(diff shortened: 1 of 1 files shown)


mauricioac:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zHboyypr&#x2F;1170-qae16imp-question-c2-2-innovation-financial-year-end-dates"> [QAE16IMP] Question C2.2 - Innovation - financial year end dates</a>

app/form_pointers/financial_table.rb
```diff
@@ -41,15 +41,23 @@ def correct_date_headers(res)
...
         entry += year.to_s
-        entry
+        fix_day_month_format(entry)
       else
-        entry
+        fix_day_month_format(entry)
       end
     end
 
     corrected_result
```
theharq: Since this file doesn't have any spec I made some debugging and found that this variable `corrected_result` is an array of string dates.. like:

```
["31/10/2010", "31/10/2011", "31/10/2012", "31/3/2013", "31/10/2014", "25/03/2016"]
```

So since we expect these dates to be correct we could just correct their format like:

```
corrected_result.map{|string_date| Date.parse(string_date).strftime("%d/%m/%Y")}
```
This will return an array:
```
["31/10/2010", "31/10/2011", "31/10/2012", "31/03/2013", "31/10/2014", "25/03/2016"]
```

So no need to  add a new method, what do you think about changing the format this way?

mauricioac: Just tested this change on irb, didn't know Date would parse dates without "0" padding.
Will push a commit in a bit

PR #1096, opened 2016-09-07
Allow superadmins edit application

theharq:
#### What this PR does:

Adds a new flag to the Admin model called superuser, this flag is meant to be set in through the console.

These admins will have the ability to edit and submit an application.

admin

#### Notes

We have updated the papertrail method `user_for_paper_trail` so we can log the admin to papertrail if he is the one making the changes


theharq:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;xEKvFk8q&#x2F;1206-qae16imp-allow-admin-to-be-able-to-edit-and-submit-an-application"> [QAE16IMP] Allow admin to be able to edit and submit an application</a>

rusllonrails:
@theharq looks fine, but if I'm not already logged in as user in same browser system redirects me to users/sign_in url, which means that I have no user session.

Probably your edit method needs one more line:
```
def edit
   authorize resource
   sign_in(@form_answer.user, bypass: true) # THIS LINE
   session[:admin_in_read_only_mode] = false
   
   redirect_to edit_form_path(resource)
end
```

rusllonrails:
@theharq also then I add this line to edit method:
```
def edit
   authorize resource
   sign_in(@form_answer.user, bypass: true) # THIS LINE
   session[:admin_in_read_only_mode] = false

   redirect_to edit_form_path(resource)
end
```
It still doesn't seem to work, I rechecked my admin on local -> and it have superadmin:true,
but I still see read-only view. tested multiple times.

rusllonrails:
Looks good :+1: 
@theharq you can merge it once Circle CI check specs

PR #1100, opened 2016-09-08
[WIP] Display audit log in admin area

theharq:
This PR adds a section to see all the version a `FormAnswer` ( Application ) has.

We are using the versions that the gem `papertrail` creates, now is possible this is not the best solution because we will have a change/version everytime the record is updated.

We cannot display where is the change was made in the form because this information is stored in a `hstore` attribute called `document` so it will show all the old json vs the new one instead of only the changes. This is for the fields in the document.

For now we are displaying when the form answer was created and updated.

logs

Any feedback will be appreciated.

### DIFF
--- a/app/controllers/admin/form_answers_controller.rb
+++ b/app/controllers/admin/form_answers_controller.rb
@@ -8,6 +8,7 @@ class Admin::FormAnswersController < Admin::BaseController
     :update_financials,
     :remove_audit_certificate
   ]
+  before_action :load_versions, only: :show
 
   skip_after_action :verify_authorized, only: [:awarded_trade_applications]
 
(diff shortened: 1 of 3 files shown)


theharq:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Ouq551WN&#x2F;1207-qae16imp-display-audit-log-for-admin"> [QAE16IMP] Display audit log for admin</a>

rusllonrails:
Looks ok :+1:

rusllonrails:
@theharq just note: there can be a lot of versions per each application which can add solid time to loading of page, so probably one of possible improvements can be rendering of versions block only if Admin clicks on 'Audit Log' arrow via AJAX

PR #1119, opened 2016-10-05
fix versions causing 404 error if changes were made by deleted user

dreamfall:


dreamfall:
[[QAE16IMP] 404 error caused by deleting an Admin that has edited/updated an application](https://trello.com/c/qti3Q1P0/1285-qae16imp-404-error-caused-by-deleting-an-admin-that-has-edited-updated-an-application)

PR #1129, opened 2016-10-24
don't filter user applications by state on post submission partial

dreamfall:
https://trello.com/c/ivgpEKD5/1227-qae16imp-after-submission-deadline-applicants-should-still-be-able-to-view-their-application


rusllonrails:
Looks ok :+1:

PR #1174, opened 2017-02-28
Popup to confirm changes when changing deadlines

mauricioac:
In the settings page, now whenever a form is "submitted", we prevent it from submitting if dates have changed and display a modal to confirm the changes.

mant8vpadf

### DIFF
--- a/app/assets/javascripts/admin/settings.js.coffee
+++ b/app/assets/javascripts/admin/settings.js.coffee
@@ -1,9 +1,90 @@
 # require jquery-ui
+
+class DeadlineForm
+  constructor: (el) ->
+    @form = el
+    @hasChanged = false
+    @init()
+
+  init: ->
+    @createModal()
+    @gatherInitialValues()
+    @bindEvents()
+
+  createModal: ->
+    html = """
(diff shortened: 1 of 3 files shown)


mauricioac:
[As an Admin I want a confirmation pop-up when changing the dates in the settings, so that I don't accidentally make changes on production](https://trello.com/c/DLlqAU5F/1562-as-an-admin-i-want-a-confirmation-pop-up-when-changing-the-dates-in-the-settings-so-that-i-don-t-accidentally-make-changes-on-pr)

app/assets/javascripts/admin/settings.js.coffee
```diff
@@ -1,10 +1,91 @@
...
+    @init()
+
+  init: ->
+    @createModal()
+    @gatherInitialValues()
+    @bindEvents()
+
+  createModal: ->
+    html = """
+<div class="modal fade" tabindex="-1" role="dialog" aria-hidden="true">
```
rusllonrails: Just for future: Looks fine - but probably also can be done using Templater like (Handlebars.js or smth)

rusllonrails:
Good :tada:

PR #1187, opened 2017-03-28
add warning copy to feedback

dreamfall:


dreamfall:
[[QAE17IMP] As an Applicant I want to see help copy in the feedback section informing me that this is not support, so that I don't request support in the wrong place](https://trello.com/c/GVkqSlfl/556-qae17imp-as-an-applicant-i-want-to-see-help-copy-in-the-feedback-section-informing-me-that-this-is-not-support-so-that-i-don-t-r)

PR #1192, opened 2017-03-30
require current password on password update

dreamfall:
https://trello.com/c/qFMVXj7V/569-as-a-user-i-should-have-to-input-my-current-password-before-i-can-create-a-new-one-so-that-my-authenticity-is-verified


rusllonrails:
good :+1:

PR #1198, opened 2017-04-04
clean infected files

dreamfall:
When an uploaded file is scanned and detected as a virus it is removed from the server immediately, preventing other users from being able to download it and preventing it from being executed on the server itself.

### DIFF
--- a/app/models/audit_certificate.rb
+++ b/app/models/audit_certificate.rb
@@ -1,7 +1,9 @@
 class AuditCertificate < ActiveRecord::Base
-
   mount_uploader :attachment, AuditCertificateUploader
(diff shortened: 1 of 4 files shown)


dreamfall:
[[QAE17IMP] As an Admin I should not be able to download uploaded files that have been flagged as a virus, so I can't be infected by it](https://trello.com/c/lX5Q8V5q/570-qae17imp-as-an-admin-i-should-not-be-able-to-download-uploaded-files-that-have-been-flagged-as-a-virus-so-i-can-t-be-infected-by)

rusllonrails:
Good job :+1:

PR #1199, opened 2017-04-04
populate feedback with submitted primary appraisal

dreamfall:
1) Only populate when the appraisals are confirmed/submitted by lead assessor
2) If they get unsubmitted, edited, resubmitted - then update the feedback accordignly
3) If RAG is G, populate Key Strength, else populate Information to strengthen the application
4) Overall verdict goes to Overall Summary
5) Add extra help copy to the section header


rusllonrails:
Looks good :+1:

PR #1200, opened 2017-04-04
add comments counter

dreamfall:
Indicate how many comments have been added on the following sections:

- Admin comments
- Critical comments


dreamfall:
[[QAE17IMP] As an Assessor I want to see the number of notes and comments that have been made, so I don't have to expand each of these sections to check if there have been any new ones added](https://trello.com/c/hAXMlwdz/470-qae17imp-as-an-assessor-i-want-to-see-the-number-of-notes-and-comments-that-have-been-made-so-i-don-t-have-to-expand-each-of-the)

PR #1201, opened 2017-04-04
hide user links for admins

dreamfall:
As an Admin/Assessor I don't want to be able to see or interact with the main menu CTAs when I view a user's application online, so I don't accidentally try and access a user's account when I am trying to access my own account


rusllonrails:
Good :+1:

PR #1202, opened 2017-04-06
Making sortable table heading arrow always visible

mauricioac:
Before it was only visible on hover

o9k2puorct


mauricioac:
[[QAE17IMP] As an Assessor I want the sort drop down arrows on the applications page to be visible all the time, not just when my cursor hovers over them, so it is obvious that they are sortable fields](https://trello.com/c/K4EYRmVR/474-as-an-assessor-i-want-the-sort-drop-down-arrows-on-the-applications-page-to-be-visible-all-the-time-not-just-when-my-cursor-hove)

PR #1205, opened 2017-04-07
hide user links for both admin and assessor

dreamfall:
display sign out link for user even if registration is not completed

https://trello.com/c/ukSESHae/539-qae17imp-as-an-admin-assessor-i-don-t-want-to-be-able-to-see-or-interact-with-the-main-menu-ctas-when-i-view-a-user-s-applicatio

### DIFF
--- a/app/views/layouts/application.html.slim
+++ b/app/views/layouts/application.html.slim
@@ -56,18 +56,17 @@
         nav#proposition-menu
           a#proposition-name href="/"
             ' Queen's Awards for Enterprise
-          - if current_admin || current_user || current_assessor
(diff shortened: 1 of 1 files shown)


dreamfall:
[[QAE17IMP] As an Admin/Assessor I don't want to be able to see or interact with the main menu CTAs when I view a user's application online, so I don't accidentally try and access a user's account when I am trying to access my own account](https://trello.com/c/ukSESHae/539-qae17imp-as-an-admin-assessor-i-don-t-want-to-be-able-to-see-or-interact-with-the-main-menu-ctas-when-i-view-a-user-s-applicatio)

PR #1208, opened 2017-04-10
Updating designs to match new designs by GDS

mauricioac:
- Updated alpha/beta colors
- Updated checkboxes and radio buttons throughout USER flows.
- Due to changing Simple Form markup style, admin and assessors had to
tweak styling in order to make the checkboxes look the same as before
- Changed buttons to have default button size of 19px, thus fixing the issue on the new collaborators page, where the button was on 14px.

The change in the markup of Simple Form is due to the fact that we were
using the nested style, which puts the checkboxes and radio buttons
inside a label along with the accompanying text.

But to make the checkboxes/radio buttons as GDS designed, they need to
be before a label with the text inside. Therefore, affecting the default
config, affected admin and assessor screens, making the checkboxes look
outside the containers.

NOTE: had to make checkboxes and radio buttons smaller than GDS' because we have a smaller font in place, so the inputs would look disproportionally big.

screen shot 2017-04-10 at 08 46 41
screen shot 2017-04-10 at 08 46 46
screen shot 2017-04-10 at 08 46 58

### DIFF
--- a/app/assets/stylesheets/_mixins.scss
+++ b/app/assets/stylesheets/_mixins.scss
@@ -94,11 +94,11 @@
     $ipad-height: 1024px;
 
     @mixin is-ios {
-      @media only screen and (min-device-width : $iphone-width) and (max-device-width : $iphone-old-height) and (orientation : landscape), 
-      only screen and (min-device-height : $iphone-width) and (max-device-height : $iphone-old-height) and (orientation : portrait), 
-      only screen and (min-device-width : $iphone-width) and (max-device-width : $iphone-height) and (orientation : landscape), 
-      only screen and (min-device-height : $iphone-width) and (max-device-height : $iphone-height) and (orientation : portrait), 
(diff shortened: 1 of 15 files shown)


mauricioac:
[[QAE17IMP] As an Applicant I want the UI styles to be updated in line with the latest GDS designs, so I know that the QAE is an official Government Digital Service](https://trello.com/c/TJX7nc13/575-qae17imp-as-an-applicant-i-want-the-ui-styles-to-be-updated-in-line-with-the-latest-gds-designs-so-i-know-that-the-qae-is-an-off)

mauricioac:
[[QAE17IMP] As a user, I want to see updated Beta site colours, so that it is in line with GDS accessibility standards](https://trello.com/c/zeCinqB9/622-qae17imp-as-a-user-i-want-to-see-updated-beta-site-colours-so-that-it-is-in-line-with-gds-accessibility-standards)

dreamfall:
👍

PR #1209, opened 2017-04-12
fix form answers sorting by flag

dreamfall:
As an Assessor/Admin I want the sorting by flag column on the applications page to function correctly, so that I can sort by applications that have been flagged


dreamfall:
[[QAE17IMP] As an Assessor I want the sorting by flag column on the applications page to function correctly, so that I can sort by applications that have been flagged](https://trello.com/c/Mz0lY4gs/624-qae17imp-as-an-assessor-i-want-the-sorting-by-flag-column-on-the-applications-page-to-function-correctly-so-that-i-can-sort-by-a)

PR #1210, opened 2017-04-18
hide enterprise promotion for admin/assessors for year >= 2017

dreamfall:
As Enterprise promotion is no longer being used from this year forward - it would be good to remove enterprise promotion from the dashboard and applications section.


dreamfall:
[[QAE17IMP] As an Admin I only want to see Enterprise Promotion results in the dashboard when application year 15/16 is selected, so I don't see unnecessary information](https://trello.com/c/Ol6DZNom/455-qae17imp-as-an-admin-i-only-want-to-see-enterprise-promotion-results-in-the-dashboard-when-application-year-15-16-is-selected-so)

app/views/admin/dashboard/index.html.slim
```diff
@@ -19,91 +19,3 @@ br
             a data-toggle="collapse" data-parent="#dashboard-details" href="#stage-one-application" aria-expanded="true" aria-controls="stage-one-application"
               ' Stage 1 - Application
         = render "admin/dashboard/stage_one_application"
-
-      /.panel.panel-default
```
rusllonrails: hey @dreamfall 
Just to confirm: we are no longer need this block?

dreamfall: @rusllonrails it was commented for ages, we don't need it

PR #1221, opened 2017-04-24
move all smtp settings into env variables

dreamfall:
Move all smtp settings into env variables to use mailtrap for development and staging server


dreamfall:
[[QAE17] We are receiving high email bounce rate notifications from AWS, we suspect invalid emails in dev and staging probably contributing to this](https://trello.com/c/PH0xYjd2/1652-qae17-we-are-receiving-high-email-bounce-rate-notifications-from-aws-we-suspect-invalid-emails-in-dev-and-staging-probably-contr)

PR #1234, opened 2017-05-08
move award year to the scope

dreamfall:
Move award year to it's own namespace, so form updates for each new award year don't affect previous award years applications.


dreamfall:
[Tech debt: Each award year should have it's own class, so form updates for each new award year don't affect previous award years applications](https://trello.com/c/RBZqjOyf/634-tech-debt-each-award-year-should-have-it-s-own-class-so-form-updates-for-each-new-award-year-don-t-affect-previous-award-years-a)

PR #1251, opened 2017-06-07
increase user timeout limit to 24h to avoid data being lost

dreamfall:
when user is considered inactive for 30 minutes


dreamfall:
@matthewford we're overriding class method that uses the config https://github.com/plataformatec/devise/blob/master/lib/devise/models/timeoutable.rb

PR #1252, opened 2017-06-09
change timeout for users to 60 minutes

dreamfall:
reload the form if user was signed out


dreamfall:
[[QAE17] As a user, I want my page to automatically refresh after 30 minutes of inactivity, so that I get redirected to the sign in page](https://trello.com/c/pUZ2xC2H/651-qae17-as-a-user-i-want-my-page-to-automatically-refresh-after-30-minutes-of-inactivity-so-that-i-get-redirected-to-the-sign-in-p)

PR #1255, opened 2017-06-26
change eligibility critirea for 3 year awards

dreamfall:
allow people who won an award from 2 years prior, rather than 3 years prior

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -702,22 +702,22 @@ jQuery ->
 
       last_year_int = parseInt(last_year)
       value_int = parseInt($(this).val())
-      not_eligible_years = [last_year_int, last_year_int - 1]
+      not_eligible_years = [last_year_int]
       not_eligible_block = lyInfo.find(".application-notice")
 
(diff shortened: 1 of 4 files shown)


dreamfall:
[[QAE17] As an applicant who is applying for a 3 year International Trade award, I should only be excluded if I was a past winner in the previous year, so that my growth in the last year can be included](https://trello.com/c/zw5GUUsU/1770-qae17-as-an-applicant-who-is-applying-for-a-3-year-international-trade-award-i-should-only-be-excluded-if-i-was-a-past-winner-in)

app/models/eligibility/trade/not_winner_in_last_two_years_validator.rb
```diff
@@ -6,9 +6,8 @@ def validate_each(record, attribute, value)
     ) unless self.class.valid_year?(value)
   end
 
-  ##
-  # must not be winner of last 2 years
+  # must not be winner last year
```
alexey-klimuk: @dreamfall It would be also good to rename validator class

dreamfall: Good point.

PR #1261, opened 2017-07-13
Trying with bundle back

mauricioac:
Now that we rebuilt without cache, maybe putting back bundle will help


rusllonrails:
@mauricioac looks fine :+1: Let's see if deploy now will pass

PR #1279, opened 2017-08-04
[CKEDITOR] set allowedContent to block extra attributes from internal sources

rusllonrails:
[TRELLO STORY](https://trello.com/c/m6hMf69b/661-qae17imp-two-applicants-are-having-issues-downloading-their-application-forms)


rusllonrails:
It's not actual - already merger in another PR

PR #1287, opened 2017-08-08
[PDF GENERATION] Promoting opportunity award - removed extra hints

rusllonrails:
[TRELLO CARD](https://trello.com/c/v0dlowsu/1826-qae17imp-promoting-opportunity-award-copy-in-section-c-on-the-downloadable-application-form-does-not-match-the-actual-applicatio)

### DIFF
--- a/app/forms/award_years/v2018/social_mobility/social_mobility_step3.rb
+++ b/app/forms/award_years/v2018/social_mobility/social_mobility_step3.rb
@@ -52,8 +52,6 @@ class AwardYears::V2018::QAEForms
           type :date
           label ->(y) { "Financial year #{y}" }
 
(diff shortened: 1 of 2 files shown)


app/forms/award_years/v2018/social_mobility/social_mobility_step3.rb
```diff
@@ -52,8 +52,6 @@ def mobility_step3
           type :date
           label ->(y) { "Financial year #{y}" }
 
-          additional_pdf_context I18n.t("pdf_texts.mobility.years_question_additional_context")
```
matthewford: We should also remove the translation string- is this used anywhere else?

rusllonrails: @matthewford yeah, you right - will remove :+1:

PR #1319, opened 2017-09-11
As an assessor, if I go into a PO assessment and then press ‘back to applications list’ the system takes me back to the SD applications list and not the PO applications list

rusllonrails:
[TRELLO CARD](https://trello.com/c/0hfGrWlD/684-qae17-as-an-assessor-if-i-go-into-a-po-assessment-and-then-press-back-to-applications-list-the-system-takes-me-back-to-the-sd-ap)


rusllonrails:
Out of date. Closing

PR #1325, opened 2017-09-13
[MANUAL FROM HARD COPY PDF GENERATION] need to use latest document

rusllonrails:
**NOW WE HAVE ISSUE:**

Now if Admin manually submits application after deadline OR
developer does manual submissing via rake script, THEN:

System generates hard copy PDF based on version of `document`
at the moment before submission end deadline.

It is wrong, because applications which Admin OR
developer submits manually after deadline can be corrected
after submission end deadline too!

**WHAT DOES THIS PR?**

1) We extended `HardCopyPdfGenerators::FormDataWorker` sidekiq worker to
   handle `generate_from_latest_doc` option, so that now we can use it for both cases:

   - generation of hard copy PDF from version of document before submission end deadline

   - and generate PDF from latest version at the moment of manul submission by Admin or Developer

2) Corrected manual submission class

3) Corrected using of `HardCopyPdfGenerators::FormDataWorker` in FormController in submission block

[TRELLO STORY](https://trello.com/c/KdWqEj8V/687-qae17-as-a-user-who-missed-the-submission-deadline-who-therefore-had-their-application-submitted-manually-by-bit-zesty-or-a-supe)

### DIFF
--- a/app/pdf_generators/hard_copy_generators/form_data_generator.rb
+++ b/app/pdf_generators/hard_copy_generators/form_data_generator.rb
@@ -7,8 +7,13 @@
 
 class HardCopyGenerators::FormDataGenerator < HardCopyGenerators::Base
   def set_pdf!
-    @pdf = form_answer.original_form_answer
-                      .decorate
+    form_record = if use_latest_version.present?
+      form_answer
+    else
+      form_answer.original_form_answer
+    end
+
+    @pdf = form_record.decorate
                       .pdf_generator
   end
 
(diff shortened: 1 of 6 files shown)


app/pdf_generators/hard_copy_generators/form_data_generator.rb
```diff
@@ -7,8 +7,13 @@
 
 class HardCopyGenerators::FormDataGenerator < HardCopyGenerators::Base
   def set_pdf!
-    @pdf = form_answer.original_form_answer
-                      .decorate
+    form_record = if use_latest_version.present?
```
dreamfall: You don't need to check if `true/false` value is present. `if use_latest_version` is enough.

PR #1340, opened 2017-09-29
Fixing autosave bug in non-editable form

mauricioac:
When assessors were going through the application, whenever they changed
steps, the autosave functionality was being called, and when that
happens and FAILS, the behaviour implemented is to refresh the page.

The fix was to simply check if the page was in read-only mode, and use
the same autosave disabling mechanism as when there's multiple
collaborators.

PS: it was believed this issue was only happening on IE, but it actually affects all browsers, as it's an issue of autosave being called, when it shouldn't

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -428,6 +428,10 @@ jQuery ->
       #
       autosave_enabled = false
 
+    # Assessor viewing form, autosave should be disabled
+    if $(".page-read-only-form").length > 0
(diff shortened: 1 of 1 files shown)


mauricioac:
[[QAE17] As an assessor viewing an application online on Internet Explorer, the page is automatically refreshing and taking me back to the top of the page](https://trello.com/c/LuNX8RnM/1935-qae17-as-an-assessor-viewing-an-application-online-on-internet-explorer-the-page-is-automatically-refreshing-and-taking-me-back)

PR #1372, opened 2017-11-22
remove duplicate of save for assessor assignments

dreamfall:


dreamfall:
[[QAE17] Assessors reported their work is not saving even though they are pressing 'save' on sections](https://trello.com/c/Ua9HG0HG/716-qae17-assessors-reported-their-work-is-not-saving-even-though-they-are-pressing-save-on-sections)

PR #1378, opened 2017-12-14
Accessibility fixes

mauricioac:
This PR accomplishes the following:

- Changes labels in questions for `legend` when the question either:
    - doesn't have any fields (just text information)
    - has checkboxes or radio buttons
    - has multiple subfields which are grouped and have their own individual labels
- Adds better labels to be listened to by screen readers
    - Buttons with "+" for example
    - Remove links without any context of which item is going to be removed
    - Question titles which were not being read in an understandable, fluid setence
- Grouping or fields belonging to one question
- Fixes some  JS issues
    - Expandable content were making the page scroll all the way up when clicked
    - Buttons to add rows (award, subsidiaries, roles, etc) were also making the page jump
- Fixes colour contrast in multiple places where it was an issue for WCAG 2.0 level AA
- Changes select2 library to NOT use `b` tag as its up/down arrow
- Skip to main content fixed. Was not the first focusable element in page, and it should be
- Fixes focus ring on logo and QAE title
- Remove extra textarea that was placed hidden


mauricioac:
Need to fix one issue before I re-open.

app/forms/qae_form_builder/question.rb
```diff
@@ -101,6 +101,22 @@ def hash_key options = {}
       options[:suffix] ? "#{delegate_obj.key}_#{options[:suffix]}" : delegate_obj.key
     end
 
+    def label_as_legend?
+      type = delegate_obj.class.name.split('::').last.gsub!(/(.)([A-Z])/,'\1_\2').downcase
+
```
dreamfall: `delegate_obj.class.name.split('::').last.underscore`

```
irb(main):001:0> "AnimalFarm".underscore
=> "animal_farm"
```

Not sure but i guess this is what you're looking for?

app/assets/javascripts/application.js.coffee
```diff
@@ -21,6 +21,14 @@
 #= require_tree ./frontend
 #= require offline
 
+ordinal = (n) ->
+  switch n
+    when 1 then n + "st"
+    when 2 then n + "nd"
+    when 3 then n + "rd"
+    else n + "th"
```
dreamfall: Won't work with all the numbers.

app/views/qae_form/_question.html.slim
```diff
@@ -6,27 +6,38 @@
...
+              span.todo
+                = ref.to_s
+          - unless question.title.blank?
+            h2
+              == question.title
 
-          /- unless question.required
-            '  (optional)
+            /- unless question.required
+              '  (optional)
```
dreamfall: Please, remove commented out code

app/views/qae_form/_question.html.slim
```diff
@@ -6,27 +6,38 @@
...
+            span class="steps step-#{ref.to_s.parameterize} #{'if-js-hide' if question.sub_ref && !question.display_sub_ref_on_js_form}"
+              span.todo
+                = ref.to_s
+          - unless question.title.blank?
+            h2
+              == question.title
 
-          == question.title
+            /- unless question.required
+              '  (optional)
```
dreamfall: Please, remove commented out code

app/views/qae_form/_question.html.slim
```diff
@@ -53,12 +64,12 @@ fieldset class=question.fieldset_classes data=question.fieldset_data_hash class=
...
-          span.visuallyhidden
-            = question.title
-            /- unless question.required
-              '  (optional)
+      / - if question.title != "" && !question.show_ref_always.present?
+      /   label for="q_#{question.key}" 
+      /     span.visuallyhidden
+      /       = question.title
+      /       /- unless question.required
+      /         '  (optional)
```
dreamfall: Please, remove commented out code

app/forms/qae_form_builder/question.rb
mauricioac: Yeah, that's probably it. I got that snippet from existing view in QAE.
But I will take your suggestion. Thanks

app/assets/javascripts/application.js.coffee
mauricioac: Nice catch! thanks

PR #1396, opened 2018-03-23
update copy & date used in email to the head of the winning organisation

dreamfall:


dreamfall:
[[QAE17] Update the Winner and Unsuccessful emails](https://trello.com/c/nqINQ9QF/764-qae17-update-the-winner-and-unsuccessful-emails)

PR #1423, opened 2018-05-02
Updating 2019's Innovation Award

mauricioac:
Copy changes for innovation award 2019, as per doc attached in this PR.

[Innovation for dev.pdf](https://github.com/bitzesty/qae/files/1968611/Innovation.for.dev.pdf)

PR #1424, opened 2018-05-03
Award financials changes

mauricioac:
Changing notice before C1 on all awards to:

> You can provide estimated figures for now but, should you be shortlisted, you will have to provide the actual figures that have been verified by an independent accountant by November.

To be merged AFTER #1423


mauricioac:
[[QAE18IMP] As an applicant, I want to see an improved help copy about financials and the use of estimates, so that I understand what information I must provide (in all award applications) (1.5)](https://trello.com/c/BEJPTEqM/701-qae18imp-as-an-applicant-i-want-to-see-an-improved-help-copy-about-financials-and-the-use-of-estimates-so-that-i-understand-what)

PR #1435, opened 2018-05-17
add PO option to A6.1 question

dreamfall:


dreamfall:
[[QAE18IMP] As an applicant, I want two new questions in section A asking if I have applied before, so that Assessors can easily access this information (6)](https://trello.com/c/RdNRNUcL/761-qae18imp-as-an-applicant-i-want-two-new-questions-in-section-a-asking-if-i-have-applied-before-so-that-assessors-can-easily-acce)

PR #1451, opened 2018-06-20
Increase password strength

tspaulino:
Related story:
 - https://trello.com/c/Bflej0qq/725-qae18-51-review-password-strength-checking-3

### DIFF
--- a/config/initializers/devise.rb
+++ b/config/initializers/devise.rb
@@ -215,5 +215,5 @@ Devise.setup do |config|
   # config.authy_remember_device = 1.month
 
   # default devise_zxcvbn minimum password score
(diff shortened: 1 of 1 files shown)


matthewford:
@tspaulino did you also check the min score in the JS front end checker?

rusllonrails:
@tspaulino ahh yeah, Matt is right.
You also need to update password score in https://github.com/bitzesty/qae/blob/master/app/assets/javascripts/frontend/password-strength-indicator.js#L105

tspaulino:
@matthewford @dreamfall @rusllonrails Changes applied. Thanks for the review 👍

PR #1453, opened 2018-06-20
Changes on collaborator sign up process

tspaulino:
Related story:
  - https://trello.com/c/V0LQSkhV/724-qae18-56-change-the-collaborator-joining-flow-6

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -69,7 +69,7 @@ class AccountsController < ApplicationController
   def update_password_settings
     current_user.set_step(5)
 
-    if (current_user.encrypted_password.blank? ? current_user.update_attributes(password_settings_params) : current_user.update_with_password(password_settings_params))
+    if update_password
       flash.notice = 'Your account details were successfully saved'
       redirect_to dashboard_path
     else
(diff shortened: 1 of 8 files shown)


app/controllers/users/confirmations_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  # def show
+  #   super
+  # end
+
+  # protected
+
+  # The path used after resending confirmation instructions.
+  # def after_resending_confirmation_instructions_path_for(resource_name)
+  #   super(resource_name)
+  # end
```
rusllonrails: @tspaulino 

I think would be good to remove all commented code from this controller, so that you will have just:
```
class Users::ConfirmationsController < Devise::ConfirmationsController

  def after_confirmation_path_for(resource_name, resource)
    if resource.is_a?(User) && resource.encrypted_password.blank?
      sign_in(resource, scope: resource_name)
        password_settings_account_path
      else
        super(resource_name, resource)
      end
    end
  end
end
```

app/controllers/accounts_controller.rb
```diff
@@ -68,11 +68,13 @@ def complete_registration
 
   def update_password_settings
     current_user.set_step(5)
-    if current_user.update_with_password(password_settings_params)
+
+    if (current_user.encrypted_password.blank? ? current_user.update_attributes(password_settings_params) : current_user.update_with_password(password_settings_params))
```
dreamfall: @tspaulino having a condition inside a condition seems a bit confusing and messy. Perhaps extract it from `if`.

tspaulino:
@dreamfall @rusllonrails Changes applied. Thanks for the review 👍

PR #1469, opened 2018-08-07
Security: Timing Attack affecting rack-protection gem

mechanicles:
- Upgrade `rack-protection` gem version to fix timing attack
vulnerability issue.
- Currently, I have upgraded the version to "1.5.5" as this also fixes the same issue.

https://trello.com/c/CbleNB3C/712-qae18-security-timing-attack-affecting-rack-protection-gem

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -87,6 +87,7 @@ gem 'vigilion-rails', '~> 1.0.5'
 gem "sidekiq", "~> 4.1.1"
 gem "sidekiq-cron", "~> 0.4.2"
 gem 'sinatra', require: nil
+gem "rack-protection", "1.5.5" # Sinatra's dependency
 
 # Redis
 gem 'redis-rails'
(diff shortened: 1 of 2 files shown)


mechanicles:
@matthewford @tomopey Security/snyk check got failed due to this **XML External Entity (XXE) Injection** issue in Nokogiri gem. I went through it, and I found that there is no proper solution until now. You can see this doc https://app.snyk.io/vuln/SNYK-RUBY-NOKOGIRI-20299, and it has this following info,

`
Nokogiri suggests not to opt-out of NONET unless only trusted documents are being parsed. There currently is no fix in libxml2 as of September 17th, 2017. Nokogiri will be waiting for a fix upstream to update.
`

You can also check this issue for the refer https://github.com/sparklemotion/nokogiri/issues/1582

I feel that we should create a new card for this Nokogiri issue on the Trello. And let's merge this issue as this issue has nothing do with Nokogiri.

PR #1476, opened 2018-08-16
Add winner notification dates on confirm email properly set based on Settings

tspaulino:
Related story:
 - https://trello.com/c/MvqkpBcJ/784-qae18-the-dates-in-the-what-happens-next-section-on-the-confirmation-of-submission-page-need-updating-3


app/models/settings.rb
```diff
@@ -35,6 +35,13 @@ def current_submission_deadline
       end
     end
 
+    def winner_notification_date
+      Rails.cache.fetch("winners_notification", expires_in: 1.minute) do
+        notification_date = current.winners_email_notification.trigger_at
```
rusllonrails: Here are are calling method `trigger_at` for notification:
```
notification_date = current.winners_email_notification.trigger_at
```
So local `notification_date` variable probably would be `DateTime` time value.

And then we are calling `trigger_at` on it again in second line:
```
notification_date.present? ? notification_date.trigger_at : nil
```
I think it will raise an exception here:
`undefined method 'trigger_at' for <DateTime...`

Also, you can use `try` instead of `a.present? ? a : nil` check:
```
def winner_notification_date
    Rails.cache.fetch("winners_notification", expires_in: 1.minute) do
       current.winners_email_notification.try(:trigger_at)
    end
end
```

app/views/qae_form/confirm.html.slim
```diff
@@ -1,3 +1,10 @@
+- winner_date = Settings.winner_notification_date
+- if winner_date
+  - month = winner_date.strftime("%B")
+  - day = winner_date.strftime("%d")
+  - ordinal_sufix = ActiveSupport::Inflector.ordinal(day)
+  - notification_date = "#{month} #{day}#{ordinal_sufix}"
```
rusllonrails: We are settings 'notification_date' local variable `if winner_date` present only.
That means that in case if `winner_date.blank?` then would be raised an issue:
```
Undefined method `notification_date` for render ...
```
You can reproduce case I mentioned if you will stub winner_date with `nil` value:
```
winner_date = nil
if winner_date
```
You probably need to specify `notification_date` with nil value for `else` case too.

dreamfall:
[[QAE18] The dates in the "what happens next" section on the confirmation of submission page need updating (3)](https://trello.com/c/MvqkpBcJ/784-qae18-the-dates-in-the-what-happens-next-section-on-the-confirmation-of-submission-page-need-updating-3)

app/views/qae_form/confirm.html.slim
dreamfall: @rusllonrails this might be a problem in many languages, but it's fine in ruby:

```
irb(main):001:0> if false
irb(main):002:1> a = 1
irb(main):003:1> end
=> nil
irb(main):004:0> a
=> nil
```

PR #1483, opened 2018-09-03
fix specs on circle ci

dreamfall:


mechanicles:
Changes look good to me 👍

PR #1487, opened 2018-09-04
remove average sector growth until SIC codes are updated

dreamfall:


dreamfall:
[[QAE18IMP] As an assessor, I don't want to see percentage growth in the case summary (3)](https://trello.com/c/NxmMgt6y/706-qae18imp-as-an-assessor-i-dont-want-to-see-percentage-growth-in-the-case-summary-3)

PR #1488, opened 2018-09-04
Remove ability to delete users

dreamfall:


dreamfall:
[[QAE18IMP] As an admin, when I delete an applicant/user from the system, I don't want the associated application form deleted, so that the application form is still available on the system (6)](https://trello.com/c/Q0tPPmWq/710-qae18imp-as-an-admin-when-i-delete-an-applicant-user-from-the-system-i-dont-want-the-associated-application-form-deleted-so-that)

mechanicles:
@dreamfall Looks good 👍

PR #1489, opened 2018-09-04
add specs to ensure we don't send useless reminders to applicants

dreamfall:
Since it was already implemented i've only added test to make sure it works as we want.


dreamfall:
[[QAE18] As an admin, I want to be able to target users who haven't submitted when sending deadline reminders emails (3.5)](https://trello.com/c/BkEEqa4t/803-qae18-as-an-admin-i-want-to-be-able-to-target-users-who-havent-submitted-when-sending-deadline-reminders-emails-35)

PR #1491, opened 2018-09-04
remove Times New Roman font from application PDF

dreamfall:


dreamfall:
[[QAE18] As an assessor, I want the answer font in the printed PDF application forms to be the same as the question text, so that it looks professional (4)](https://trello.com/c/S9GtNcmf/766-qae18-as-an-assessor-i-want-the-answer-font-in-the-printed-pdf-application-forms-to-be-the-same-as-the-question-text-so-that-it)

PR #1496, opened 2018-09-07
use shared memory folder for CI to speed up tests

dreamfall:


mechanicles:
@dreamfall Can you share the doc/article for this change if there is any so that I can go through it as well.

PR #1503, opened 2018-09-07
Loosen up date validation in Q. A5 to use the latest date as a max date

dreamfall:


dreamfall:
[[QAE18] Extra date validation in questions A5 to be only the one specified in the main text (not the one in the brackets)](https://trello.com/c/fVyY1MG6/815-qae18-extra-date-validation-in-questions-a5-to-be-only-the-one-specified-in-the-main-text-not-the-one-in-the-brackets)

PR #1526, opened 2020-01-03
upgrade jquery-ui-rails

macool:
to avoid an XSS vulnerabilty


macool:
replaces #1514

macool:
replaces #1515

PR #1566, opened 2020-10-16
Create Audit Logs For Key Events

dvjones89:
One of the biggest challenges currently faced by the QAE team is keeping track of which applications have recently been updated.

This MR adds lots of new hooks into `admin/` and `assessor/` controller actions to create new `AuditLog` records.

Using these `AuditLog` records we can then populate the "Application Audit Log" panel with lots of additional information about the changes that have been made to a `FormAnswer` by `Admin` and `Assessor` users.

https://trello.com/c/YQlNYgcA/1160-qaeimp-last-updated-column-on-applications-page

Screenshot 2020-10-16 at 16 15 20


RyanMacG (approved):
LGTM once test issues are resolved

PR #1573, opened 2020-10-19
🐛 Handle `nil` values for `PaperTrail::Version.whodunnit`

dvjones89:
We've recently seen some Sentry errors trying to call `.split` on a
`nil` object. After looking closer at the issue, it seems we're
occasionally having trouble generating our list of `AuditEvent` records
due to some `PaperTrail::Version` objects having a `nil` value for
`whodunnit`.

This commit avoids the runtime error by moving our `dummy_user` helper
method out of `ApplicationController` and into a reusable `audit_helper`
module. This allows us to return a dummy user record whenever the
`PaperTrail` `whodunnit` value is `nil`.

### DIFF
--- a/app/helpers/audit_helper.rb
+++ b/app/helpers/audit_helper.rb
@@ -1,0 +1,19 @@
+module AuditHelper
+
+  def dummy_user
+    User.find_by(email: "dummy_user@bitzesty.com") || User.create!(dummy_user_params)
+  end
+
+  private
+
+  def dummy_user_params
+   {
+     email: "dummy_user@example.com",
+     password: SecureRandom.base64(16),
+     agreed_with_privacy_policy: '1',
+     role: "regular",
+     first_name: "Unknown",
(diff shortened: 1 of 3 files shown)


app/helpers/audit_helper.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  def dummy_user
+    User.find_by(email: "dummy_user@bitzesty.com") || User.create!(dummy_user_params)
+  end
+
+  private
+
+  def dummy_user_params
+   {
+     email: "dummy_user@bitzesty.com",
```
matthewford, line 11: can we use @example.com to not send us potential email.

matthewford (approved):

app/helpers/audit_helper.rb
dvjones89: Sure, changed. Thanks for the review!

PR #1600, opened 2020-10-27
📝 Tweak copy for account verification pages

dvjones89:
This MR makes some last-minute amendments to the copy displayed to shortlisted users who must upload their verification documents (Accountant's Report and list of procedures).

https://trello.com/c/UGpnY8ei/1194-qaeimp-add-an-upload-list-of-procedures-feature

screencapture-qae-dvjon-es-3000-dashboard-2020-10-27-15_47_32
screencapture-qae-dvjon-es-3000-users-form-answers-1-audit-certificate-2020-10-27-15_47_19

### DIFF
--- a/app/controllers/users/list_of_procedures_controller.rb
+++ b/app/controllers/users/list_of_procedures_controller.rb
@@ -20,7 +20,7 @@ class Users::ListOfProceduresController < Users::BaseController
       format.html do
         if saved
           redirect_to users_form_answer_audit_certificate_url(form_answer),
-                      notice: "List of Procedures successfully uploaded!"
+                      notice: "List of procedures successfully uploaded!"
         else
           render :show
         end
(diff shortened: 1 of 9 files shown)


app/controllers/users/list_of_procedures_controller.rb
```diff
@@ -20,7 +20,7 @@ def create
       format.html do
         if saved
           redirect_to users_form_answer_audit_certificate_url(form_answer),
-                      notice: "List of Procedures successfully uploaded!"
+                      notice: "list of procedures successfully uploaded!"
```
mauricioac, line 23: @dvjones89 shouldn't `list` be capitalized?

mauricioac (approved):
@dvjones89 I'll leave the PR approved in case the change I suggested should remain as is. But other than that, all good :100:

app/controllers/users/list_of_procedures_controller.rb
dvjones89: Yup, you're totally right, good catch. Fixed now 🙇

PR #1619, opened 2020-11-18
Improve UX for upload and download of verification documents

dvjones89:
As part of our ongoing efforts to improve the QAE UX, we'd like the
download links for user-uploaded documents to feature the name of the
document, making it easier for the user to verify what they've uploaded
and to understand what they're about to re-download.

hJtW9cTixT

### DIFF
--- a/app/views/users/audit_certificates/_file.html.slim
+++ b/app/views/users/audit_certificates/_file.html.slim
@@ -4,7 +4,7 @@
   - if att_record.attachment_scan_results.present?
     - if att_record.clean?
       = link_to att_record.attachment.file.filename, att_record.attachment_url,
-                target: "_blank", class: "js-audit-certificate-title download-link"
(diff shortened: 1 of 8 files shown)


app/views/users/audit_certificates/_file.html.slim
```diff
@@ -3,21 +3,21 @@
 
   - if att_record.attachment_scan_results.present?
     - if att_record.clean?
-      = link_to "Download External Accountant's Report", att_record.attachment_url,
-                target: "_blank", class: "js-audit-certificate-title"
+      = link_to att_record.attachment.file.filename, att_record.attachment_url,
```
mauricioac, line 6: @dvjones89 for the download links, you can also add a `download` attribute which will make browser force download, which also helps some browsers not retain a blank tab

mauricioac (approved):

PR #1623, opened 2020-11-18
🐛 Update how filename is extracted from URLs

dvjones89:
We've recently updated the Verification of Commercial Figures UI to
display the name of your file during upload. Extracting the filename
from the URL was achieved rather naively using a method that doesn't
play nicely when a URL contains query parameters.

This commit extracts the filename from the URL using a more specialised
regular expression, which will hopefully correctly handle URLs with
query parameters.


mauricioac (approved):
I'm approving this, but does this happen?

I have never seen a file url with a query string before

PR #1707, opened 2021-04-30
fix matrix question backend validation

dreamfall:


mauricioac (approved):
oh wow, makes total sense

PR #1780, opened 2021-09-15
Main prod sync and reminder changes

dreamfall:
https://app.asana.com/0/1200504523179343/1200979643814307


TheDancingClown (approved):

app/workers/scheduled/email_notification_service_worker.rb
```diff
@@ -1,6 +1,5 @@
 module Scheduled
   class EmailNotificationServiceWorker < BaseWorker
-    sidekiq_options retry: 0
```
matthewford, line 3: We want to keep this to prevent retry

Gemfile
```diff
@@ -4,7 +4,7 @@ git_source(:github) { |name| "https://github.com/#{name}.git" }
 
 ruby '~> 2.6.6'
 
-gem 'rails', '~> 6.0.4.1'
+gem 'rails', '~> 6.0.3.7'
```
matthewford, line 7: We should keep the rails upgrade

app/models/form_answer.rb
```diff
@@ -344,11 +344,7 @@ def unsuccessful_app?
   end
 
   def collaborators
-    if account
-      account.collaborators_with(user)
-    else
-      User.none
-    end
+    account.collaborators_with(user)
```
matthewford, line 347: Should we not keep the guard in case this issue happens again?

PR #1793, opened 2021-09-20
reverts mimemagic removal

TheDancingClown:


DaniBitZesty (approved):

matthewford:
it should be safe to remove, I think it was left in by accident after rails upgrade

PR #1855, opened 2021-10-28
Removed aria-labels from legends

DaniBitZesty:
Aria-labels are not accepted attributes on legends, and were raising the issue below on Axe: 

Screenshot 2021-10-28 at 16 38 54

"Elements must only use allowed ARIA attributes" is a MUST requirement for WCAG 2.0 (A) ([ref](https://dequeuniversity.com/rules/axe/4.3/aria-allowed-attr?application=AxeChrome)). This layout of `legend`, then `label` is replicated from the QAVS question template, where some legends were empty but used to satisfy fieldset + legend relationship. @mauricioac what are you thoughts on this solution vs reworking `_question.html slim` to remove some questions from fieldsets altogether?

UPDATED: 
- Added aria-required attribute to required questions
- Removed aria-labels from legends
- Only questions listed within `legend_types` in `qae_form_builder/question.rb` are rendered as fieldsets with legend, all other questions are within divs with the same classes
- Added a dark govuk blue to meet contrast ratio against blue background on confirmation questions
- Added govuk styling to intro text in step 4

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -241,10 +241,6 @@ jQuery ->
     $(this).prop('required', true)
     $(this).attr('aria-required', 'true')
 
-  $('.qae-form').find('input[type="number"]').each -> 
-    $(this).attr('pattern', '[0-9]*')
-    $(this).attr('inputmode', 'decimal')
-
   # Calculates the UK Sales for Sus Dev form
   # UK sales = turnover - exports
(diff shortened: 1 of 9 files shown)


mauricioac:
There should not be a aria-label on legends

legends are needed on groups of fields, like address, checkboxes and on radio buttons.
I'm not sure how easy it would be to fix this on _question.html.erb

TheDancingClown (approved):

PR #1865, opened 2021-11-03
Add aria-describedby to inputs where error message appears

DaniBitZesty:
Card: https://app.asana.com/0/1200504523179345/1200504476655848

- Assigns each error message a unique id
- Assigns each input with an error message an aria-describedby that references the error's id

Screenshot 2021-11-03 at 10 33 51

Updated:
Removes aria-describedby when error is cleared
Screenshot 2021-11-03 at 13 29 17

Adds aria-describedby when error appears
Screenshot 2021-11-03 at 13 29 26

### DIFF
--- a/app/assets/javascripts/frontend/form-validation.js.coffee
+++ b/app/assets/javascripts/frontend/form-validation.js.coffee
@@ -20,25 +20,22 @@ window.FormValidation =
 
   addErrorMessage: (question, message) ->
     @appendMessage(question, message)
+    @addAriaDescribedByToInput(question)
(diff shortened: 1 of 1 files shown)


mauricioac:
@DaniBitZesty do you remove the aria-describedby when the input has no error? cause when the input is not showing a validation error, the attribute should not point to the error message.

DaniBitZesty:
@mauricioac I had not, no... Have just added the `clearAriaDescribedby` function and updated the description in this PR with images for proof lol

mauricioac (approved):

PR #1877, opened 2021-11-05
Correcting errors on head_of_business subfields

DaniBitZesty:
- Corrected spelling of head_of_business_title from head_of_bussines_title
- Corrected spelling of head_of_business_job_title from head_job_title
- Corrected spelling of head_of_business_email from head_email
- Added reference number back into head_of_business question
- Restructured head_of_business question template


matthewford:
Probably safer to do this in the next years form I think or we will need a data migration

PR #1910, opened 2021-11-29
Toolkit upgrade + accessibility fixes

mauricioac:
Massive work to upgrade the GDS toolkit + many accessibility fixes cross-polinated from other projects.


dreamfall (changes requested):
Ok, not sure about the commented out css/js, but need to reenable the pending spec at least.

dreamfall (approved):

PR #1931, opened 2021-12-16
Hot fix on contact details styling

DaniBitZesty:
- The fields for '_Title_', '_Job title_' and '_Email address_' were styled differently to the other fields within the 'Head of business' question because they are not subquestions.
- There is a PR that converts all 'Head of business' questions into subquestions (#1877) but it involves a lot of changes, because in addition to making them into subfields, spelling and formatting had to be corrected to '_head_of_business__[field]'. (previously '_head_of_bussines_title_' or '_head_email_')

This PR provides an alternative. Excluding questions with `excluded_header_questions` attribute from having `govuk-!-font-size-24 govuk-!-font-weight-bold` classes applied by the `_question.html.erb` template.

Before:
Screenshot 2021-12-16 at 16 15 32
After:
Screenshot 2021-12-16 at 16 14 46

### DIFF
--- a/app/forms/award_years/v2022/innovation/innovation_step6.rb
+++ b/app/forms/award_years/v2022/innovation/innovation_step6.rb
@@ -9,6 +9,7 @@ class AwardYears::V2022::QAEForms
         text :head_of_bussines_title, "Title" do
           required
           classes "sub-question"
+          excluded_header_questions
           style "tiny"
         end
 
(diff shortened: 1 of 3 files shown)


DaniBitZesty:
@matthewford  just occurred to me that this award year has closed, so we may not need a temporary fix for this question's styling. It is still a simpler alternative to PR #1877.

matthewford:
Staff will view the form so probably worth adding still

TheDancingClown (approved):

PR #1940, opened 2021-12-22
🔧  fixes upload organisational chart feature

TheDancingClown:
this PR fixes some classes lost in the toolkit upgrade.
- upon adding an organisational chart file the styling was not correct. This is fixed by adding the html to the JS function adding the attachment display.
- upon adding an organisational chart file the upload button was still showing. This is fixed by adding the visually hidden class to the correct element (govuk_button instead of button).
- upon removing an organisational chart file the upload button was not showing. This is fixed by find the correct element (searching for .button_add instead of .js-button-add)

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -572,7 +572,7 @@ jQuery ->
 
     upload_started = (e, data) ->
       # Show `Uploading...`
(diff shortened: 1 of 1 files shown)


mauricioac (approved):

mauricioac:
but you do know that visually hidden means tabbable by screen readers right?

PR #2043, opened 2022-04-28
copy/style changes. change default value of the new column

dreamfall:
https://app.asana.com/0/1200504523179345/1202184759193252/f

### DIFF
--- a/app/views/palace_invites/_form.html.slim
+++ b/app/views/palace_invites/_form.html.slim
@@ -41,7 +41,11 @@
 
             = ff.input :has_royal_family_connections, as: :radio_buttons, label: false, input_html: { class: "js-has-royal-connections" }
           .question-group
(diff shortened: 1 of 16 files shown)


TheDancingClown (approved):

app/views/palace_invites/_form.html.slim
```diff
@@ -9,7 +9,7 @@
 
         li.js-add-example
           = hidden_field_tag "palace_invite[palace_attendees_attributes][#{i}][_remove]", "", class: 'remove'
-          a.remove-link.js-remove-link href="#" Remove
+          / a.remove-link.js-remove-link href="#" Remove
```
TheDancingClown, line 12: Are you leaving this comment in?

dreamfall: yes, we might need it later when number of attendees increases back to 2

TheDancingClown (approved):

PR #2073, opened 2022-05-13
Adds a confirm deletion page when deleting only collaborator

TheDancingClown:
https://app.asana.com/0/1200504523179345/1202087036595965/f

Account owners need a collaborator to be able to create and edit applications so a confirm deletion page is introduced to highlight the restrictions users may face when deleting the last collaborator.

- If there is only 1 collaborator, the 'Remove Collaborator' link takes the user to an interim page where they will confirm they wish to delete. 
- Changes the link colour to red
- Original feature tests confirm deletion with multiple collaborators, feature test added to confirm redirection when user has only one collaborator.

Screenshot 2022-05-13 at 13 08 05

### DIFF
--- a/app/views/account/collaborators/confirm_deletion.html.slim
+++ b/app/views/account/collaborators/confirm_deletion.html.slim
@@ -1,0 +1,24 @@
+- title "Confirm removal of collaborator"
+
+- provide(:page_content_class, "multi-page page-account")
+
+.article-container
+  article.group role="article"
+    h1.govuk-heading-l
+      ' Are you sure you want remove this collaborator?
+
+    p.govuk-body
+      ' You need to have at least one collaborator added to your account. If you remove this collaborator, please add another one; otherwise, you won't be able to:
+    ul.govuk-list.govuk-list--bullet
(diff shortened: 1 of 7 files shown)


app/views/account/collaborators/confirm_deletion.html.slim
```diff
@@ -0,0 +1,24 @@
...
+    p.govuk-body
+      ' You need to have at least one collaborator added to your account. If you remove this collaborator, please add another one; otherwise, you won't be able to:
+    ul.govuk-list.govuk-list--bullet
+      li access or submit your applications;
+      li access the winners' dashboard if you win the award.
+
+    .govuk-button-group
+      - form_id = params.key?(:form_id) ? params[:form_id] : nil
+
+      = link_to "Remove collaborator", account_collaborator_path(collaborator, form_id: form_id),
```
dreamfall, line 19: I think you can just pass `params[:form_id]` or `params[:form_id].presence` if you don't want a blank string

dreamfall (changes requested):

PR #2094, opened 2022-06-28
count account owner as a collaborator

dreamfall:
https://app.asana.com/0/1200504523179343/1202504961108164/f

Inviting account admin as a first collaborator let them being stuck on collaborators page because account owner was excluded from collaborators. 

This should still work fine since we exclude current user from the scope, looks like it was just a legacy code.

### DIFF
--- a/app/views/account/collaborators/index.html.slim
+++ b/app/views/account/collaborators/index.html.slim
@@ -76,7 +76,7 @@ h1.govuk-heading-xl
                   = link_to new_account_collaborator_path(form_id: @form_answer), class: "govuk-button", role: "button"
                     ' Add a collaborator
 
-              - if collaborators.any?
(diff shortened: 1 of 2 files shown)


TheDancingClown (approved):

app/views/account/collaborators/index.html.slim
```diff
@@ -76,7 +76,7 @@ h1.govuk-heading-xl
                   = link_to new_account_collaborator_path(form_id: @form_answer), class: "govuk-button", role: "button"
                     ' Add a collaborator
 
-              - if collaborators.any?
+              - if current_account.users.not_including(account_owner).any?
```
TheDancingClown, line 79:
```suggestion
              - if all_collaborators.any?
```

PR #2121, opened 2022-08-04
Adds all years param for csv reports

TheDancingClown:
https://app.asana.com/0/1200061634447616/1202697541255350

The award year can not be used when the All Years option is selected when downloading reports, as the 'all_years' param cannot be converted to an integer. Therefore the param is passed into the report builder methods and a check is done to use all FormAnswers if this param is found.

### DIFF
--- a/app/models/reports/admin_report.rb
+++ b/app/models/reports/admin_report.rb
@@ -10,17 +10,17 @@ class Reports::AdminReport
   def as_csv
     case id
     when "registered-users"
-      Reports::RegisteredUsers.new(year).build
+      Reports::RegisteredUsers.new(year, params).build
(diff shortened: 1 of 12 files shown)


DaniBitZesty (approved):

app/models/reports/admin_report.rb
```diff
@@ -10,17 +10,17 @@ def initialize(id, year, params = {})
   def as_csv
     case id
     when "registered-users"
-      Reports::RegisteredUsers.new(year).build
+      Reports::RegisteredUsers.new(year, @params).build
```
matthewford, line 13: Could be just params

PR #2144, opened 2022-10-05
Shortlisted Dashboard update

dreamfall:
https://app.asana.com/0/1200504523179345/1202154074583687

- add new mailers for mobility/PO forms
- add new file upload for mobility/po forms
- change the dashboard layout
- add ability to review new shortlisted doc type for admin/assessor


app/models/form_answer.rb
```diff
@@ -161,6 +164,9 @@ def secondary
     scope :primary_and_secondary_appraisals_are_not_match, -> {
       where("discrepancies_between_primary_and_secondary_appraisals::text <> '{}'::text")
     }
+
+    scope :require_vocf, -> { where(award_type: %w[trade innovation]) }
```
TheDancingClown, line 168: Do you also need to check the award year (pre 2023) and if busines in this scope?

app/models/shortlisted_documents_wrapper.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  private
+
+  def valid_for_submission?
+    return true unless submitted?
+
+    if vat_returns_files.first&.attachment.present?
+      true
+    else
+      errors.add(:base, "You need to add at least one VAT returns file")
```
TheDancingClown, line 35:
```suggestion
      errors.add(:base, "You need to add at least one VAT return file")
```

TheDancingClown (approved):

app/models/form_answer.rb
dreamfall: I think it's gonna be fine like that, moving forward it's only gonna be used on applications > 2023.

app/models/shortlisted_documents_wrapper.rb
dreamfall: I think it should be `VAT returns` cause the input is named `Upload VAT returns`

PR #2146, opened 2022-10-11
QA copy changes VOFF

TheDancingClown:
https://app.asana.com/0/home/1200061634431011/1203115775652532

Copy changes - see asana sub-tasks for details.

In addition to copy changes - the audit certificate upload button has been removed for PO/SD applications. This is a temporary change to assess the new changes to these award types.

### DIFF
--- a/app/renderers/mail_renderer.rb
+++ b/app/renderers/mail_renderer.rb
@@ -243,7 +243,14 @@ class MailRenderer
     d = deadline(kind)
 
     if d.present?
-      d.strftime(format)
+      deadline_time = d.strftime(format)
+      if deadline_time == "12:00"
+        "noon"
+      elsif deadline_time == "00:00"
+        "midnight"
+      else
+        deadline_time
+      end
     else
       DateTime.new(Date.current.year, 9, 21, 10, 30).strftime(format)
     end
--- a/spec/features/assessor/review_shortlisted_docs_spec.rb
+++ b/spec/features/assessor/review_shortlisted_docs_spec.rb
@@ -22,7 +22,7 @@ describe "Assessor reviews the shortlisted docs" do
       end
 
       within "#financial-summary" do
-        expect(page).to have_selector("p", text: "No change necessary")
+        expect(page).to have_selector("label", text: "No change necessary")
       end
     end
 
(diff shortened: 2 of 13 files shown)


app/renderers/mail_renderer.rb
```diff
@@ -243,7 +243,13 @@ def deadline_str(kind, format = "%d/%m/%Y")
     d = deadline(kind)
 
     if d.present?
-      d.strftime(format)
```
dreamfall, line 246: we should assign it to a var to avoid calling `strftime` 3 times most of the time.

app/views/admin/form_answers/_financials_review.html.slim
```diff
@@ -54,7 +54,8 @@
         = f.input :form_answer_id, as: :hidden
 
         .form-group
-          = render "admin/form_answers/financial_summary/shortlisted_financial_docs", docs_wrapper: resource.shortlisted_documents_wrapper
+          - if resource.requires_vocf?
+            = render "admin/form_answers/financial_summary/shortlisted_financial_docs", docs_wrapper: resource.shortlisted_documents_wrapper
```
dreamfall, line 58: this is not correct, we should display it unless `resource.requires_vocf?`. But we already check it at the top change is not needed.

spec/features/assessor/review_shortlisted_docs_spec.rb
```diff
@@ -1,3 +1,5 @@
+# This test suite has been marked as pending as the functionality for assessors to download audit certificates has been temporarily removed to see how the current system functions
```
dreamfall, line 1: This should still work

dreamfall (changes requested):

app/views/admin/form_answers/_financials_review.html.slim
TheDancingClown: I have changed the policy for downloading the commercial figures. This policy is only used in one place and we have been asked to remove the download link for assessors and admins

spec/features/assessor/review_shortlisted_docs_spec.rb
TheDancingClown: Fixed the selector

dreamfall (approved):

PR #2186, opened 2022-12-01
Fetching applications from the same entity

Lubosky:
Asana card here: https://app.asana.com/0/1200504523179345/1202115307313426/f

 - Added validator for validating company registration number. Works well with correctly formatted numbers as well as numbers having leading zero(s) omitted
 - Added extractor service class for extracting company registration numbers from content. Simple string as well as HTML block can be passed to it. It also works when extracting multiple numbers
 - Query to get other applications by the same entity based on the submitted registration number. Still need some work as for example if on the current application the registration number is `6883289`, we may want to look for the applications containing also `06883289` as that is the valid format of the same number.


dreamfall (approved):
Need to change to only display submitted applications and :shipit:

PR #2201, opened 2022-12-29
International award help text

Lubosky:
Asana card here: https://app.asana.com/0/1200504523179345/1203151450169274/f

### DIFF
--- a/app/assets/javascripts/application.js.coffee
+++ b/app/assets/javascripts/application.js.coffee
@@ -1014,6 +1014,7 @@ jQuery ->
     $(this).closest("span").toggleClass("open")
 
   OptionsWithPreselectedConditionsQuestion.init()
+  WonInternationalTradeAwardQuestion.init()
   ongoingDateDuration()
   SupportLetters.init()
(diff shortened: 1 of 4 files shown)


app/forms/award_years/v2023/international_trade/international_trade_step1.rb
```diff
@@ -247,6 +247,8 @@ def trade_step1
 
           children_options_depends_on :category
           dependable_values [:international_trade]
+
+          additional_pdf_context I18n.t("pdf_texts.trade.queen_awards_question_additional_context"), if: -> () { Current.form_answer&.won_international_trade_award_last_year? }
```
dreamfall, line 251: I think this will work. You could also just add a special method to render this context for those questions and have a check inside that method when it's rendered in pdf. Or just pass a method name with `if: :won_international_trade_award_last_year?` and call on form answer there.

I know that CurrentAttributes is thread safe, but the way this code is loaded at the start of application in production still makes me feel a bit uneasy. This should work though in theory.

dreamfall (approved):

PR #2266, opened 2023-03-09
Stream CSV reports

Lubosky:
Some admin reports were timeout-ing on CloudFoundry. Before optimizing methods it's better to try to stream CSV responses as that may solve the issue even though the CSV reports may still take 1 minute or more to generate.

Asana card here: https://app.asana.com/0/1200504523179343/1204079131266895/f

Enough data to test this is on production.

## ✅ Checklist

- [ ] I have squashed any unnecessary or part-finished commits
- [ ] I have attached screenshots of visual changes

### DIFF
--- a/app/models/reports/admin_report.rb
+++ b/app/models/reports/admin_report.rb
@@ -12,7 +12,7 @@ class Reports::AdminReport
     when "registered-users"
       Reports::RegisteredUsers.new(year).stream
     when "press-book-list"
-      Reports::PressBookList.new(year).build
+      Reports::PressBookList.new(year).stream
     when "cases-status"
       Reports::CasesStatusReport.new(year).build
     when "entries-report"
--- a/app/models/reports/admin_report.rb
+++ b/app/models/reports/admin_report.rb
@@ -16,7 +16,7 @@ class Reports::AdminReport
     when "cases-status"
       Reports::CasesStatusReport.new(year).build
     when "entries-report"
-      Reports::AllEntries.new(year).build
+      Reports::AllEntries.new(year).stream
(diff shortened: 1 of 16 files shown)


app/models/reports/admin_report.rb
```diff
@@ -10,17 +10,17 @@ def initialize(id, year, params = {})
...
+      Reports::AllEntries.new(year).stream
     when "discrepancies_between_primary_and_secondary_appraisals"
-      Reports::DiscrepanciesBetweenPrimaryAndSecondaryAppraisals.new(year, params[:category]).build
+      Reports::DiscrepanciesBetweenPrimaryAndSecondaryAppraisals.new(year, params[:category]).stream
     when "reception-buckingham-palace"
-      Reports::ReceptionBuckinghamPalaceReport.new(year).build
+      Reports::ReceptionBuckinghamPalaceReport.new(year).stream
     when /assessors-progress/
       if FormAnswer::AWARD_TYPE_FULL_NAMES.keys.include?(params[:category])
         Reports::AssessorsProgressReport.new(year, params[:category]).build
```
TheDancingClown, line 26:
```suggestion
        Reports::AssessorsProgressReport.new(year, params[:category]).stream
```

TheDancingClown: Does this also need to be changed?

dreamfall (approved):

app/models/reports/admin_report.rb
Lubosky: That one is probably better to stay for now as this class `Reports::DataPickers::AssessorProgressPicker` would have to be rewritten to lazy load the records. Was checking it briefly and for now there should not be any performance issues with this one.

PR #2381, opened 2023-05-09
Fix financial step calculation with regards to previous years

Lubosky:
Financial step changed from being the 3rd step to 4th. This caused issue with forms from previous years as they were not displaying financial summary. 
Probably would be enough to check what's the year for the form but as I was looking at the history of changes I thought the logic was more complicated so I added hash as constant which would define financial step for each year and award type. Will leave it as is for the next year or we can just go for simple condition.

Asana card here: https://app.asana.com/0/1200504523179343/1204553128463890/f

None

## ✅ Checklist


TheDancingClown (approved):
Very nice solution

PR #2474, opened 2023-08-08
Adds non-js validation for matrix cells

DaniBitZesty:

* Adds required field below empty cells when submitting a form with JS-disabled

* https://app.asana.com/0/1204861597377754/1205042279939911/f

* 

## ✅ Checklist


forms/award_years/v2024/social_mobility/social_mobility_step3.rb
```diff
@@ -543,6 +543,7 @@ def mobility_step3
           classes "sub-question question-matrix"
           ref "C 5.2"
           required
+          required_rows :initiative_activities
```
TheDancingClown, line 546: @DaniBitZesty is this correct?

forms/qae_form_builder/matrix_question.rb
```diff
@@ -8,7 +8,7 @@ def errors
           question.x_headings.each do |x_heading|
             suffix = "#{x_heading.key}_#{y_heading.key}"
 
-            if !question.input_value(suffix: suffix).present? && question.required_rows.include?(y_heading.key)
+            if !question.input_value(suffix: suffix).present?
```
TheDancingClown, line 11: This change makes all rows required, correct?

TheDancingClown: I see that it will work for non-js forms, but we still need the conditional requirement for back end validation with JS enabled? Otherwise no point in that conditional row requirement

TheDancingClown (changes requested):

forms/award_years/v2024/social_mobility/social_mobility_step3.rb
DaniBitZesty: @TheDancingClown this is the key of the question the matrix validations are dependant on. It's working correctly when I test it, what do you think should be passed in instead?

TheDancingClown (approved):

PR #2525, opened 2023-08-24
Adds timeout warning popup

TheDancingClown:

* Shows a warning popup with 5 minutes remaining to admins/assessors/judges and gives the option to extend the session
* Increases session timeout to 20 hours for applicants.

https://www.w3.org/WAI/WCAG21/Understanding/timing-adjustable

* https://app.asana.com/0/1204861597377754/1205067405053454

* None

## ✅ Checklist

## 🖼️ Screenshots (if appropriate - no PII/Prod data):

Screenshot 2023-08-29 at 11 05 30


app/models/user.rb
```diff
@@ -161,7 +161,7 @@ def new_member?
   end
 
   def timeout_in
-    60.minutes
+    1200.minutes
```
Lubosky, line 164: This would be better to put in ENV variable in case we want to quickly change it - can have 20 hrs as default value. Also better to use hours for readability.

app/views/shared/_timeout_warning_popup.html.slim
```diff
@@ -0,0 +1,25 @@
...
+          span aria-hidden="true"
+            | &times;
+
+        h4
+          | Session Expiration Warning
+
+        p
+          | In
+          span.time-target
+          = "minutes you will be automatically logged out because of prolonged inactivity (#{distance_of_time_in_words(Time.now, Time.now + current_subject.timeout_in, only: [:minutes, :hours])}), to keep your data safe."
```
Lubosky, line 15: Maybe move to locale file so you get plural right... 1 minute, 2 minutes

Lubosky (changes requested):

app/views/shared/_timeout_warning_popup.html.slim
TheDancingClown: I've done this in the javascript instead as this is where the time is added

PR #2567, opened 2023-09-04
[staging release] use function instead of new syntax for older version of uglifier

dreamfall:
addressing https://github.com/bitzesty/qae/actions/runs/6049304976/job/16417630855

### DIFF
--- a/app/assets/javascripts/frontend/custom_selects.js
+++ b/app/assets/javascripts/frontend/custom_selects.js
@@ -9,7 +9,7 @@ $(document).ready(function () {
     accessibleAutocomplete.enhanceSelectElement({
       selectElement: field,
       showAllValues: true,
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/frontend/custom_selects.js
```diff
@@ -9,7 +9,7 @@ $(document).ready(function () {
     accessibleAutocomplete.enhanceSelectElement({
       selectElement: field,
       showAllValues: true,
-      dropdownArrow: () => {
+      dropdownArrow: fucntion() {
```
TheDancingClown, line 12:
```suggestion
      dropdownArrow: function() {
```

TheDancingClown (changes requested):

TheDancingClown (approved):

PR #2579, opened 2023-09-05
Revert "Try to fix fetching form answers for case summary report"

Lubosky:
Reverts #2577


Lubosky:
Cancelling this as this was opened just facilitate the scenario where #2577 would not work. It worked and the console logs were already cleaned up.

PR #2640, opened 2023-10-18
.minmax should be called on arrays

DaniBitZesty:

* This was causing the error `undefined method strftime for nil:NilClass` 
* `(from..to).minmax` was returning [nil, nil]
* .minmax cannot be called on a range, so this converts (from..to) to [from, to] array

https://appsignal.com/bit-zesty/sites/601bea9d7478202eb7a88e16/exceptions/incidents/122/samples/last

* https://app.asana.com/0/1200504523179343/1205732709868441/f

* 

## ✅ Checklist

### DIFF
--- a/app/models/award_year.rb
+++ b/app/models/award_year.rb
@@ -314,8 +314,8 @@ class AwardYear < ApplicationRecord
       end_date = Date.new(AwardYear.current.year - 1 - from, month, day)
 
       if opts[:minmax] == true
-        return (start_date..end_date).minmax.map { |d| d.strftime("%d/%m/%Y") } if opts[:format] == true
-        (start_date..end_date).minmax
+        return [start_date, end_date].minmax.map { |d| d.strftime("%d/%m/%Y") } if opts[:format] == true
+        [start_date, end_date]
       else
         start_date..end_date
       end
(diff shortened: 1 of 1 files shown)


app/models/award_year.rb
```diff
@@ -333,8 +333,8 @@ def fy_date_range_threshold(**opts)
       to = to.change(year: year)
 
       if opts[:minmax] == true
-        return (from..to).minmax.map { |d| d.strftime("%d/%m/%Y") } if opts[:format] == true
-        (from..to).minmax
+        return [from, to].minmax.map { |d| d.strftime("%d/%m/%Y") } if opts[:format] == true
+        [from, to].minmax
```
Lubosky, line 337: I realized we don't need to call `minmax` here as only `[from, to]` would be enough

Lubosky (changes requested):
We can remove `minmax` from being called... Other than that it's ok.

app/models/award_year.rb
DaniBitZesty: @Lubosky I've also updated this in `start_trading_between` method

Lubosky (approved):
👍

PR #2746, opened 2024-03-20
Adds new CSV report for user data

TheDancingClown:

* Adds Reports::User class to generate report data for different user types. The Admin/Assessor/Judge user types are separate tables and have different columns - where a column does not exist "N/A" is returned.
* Adds roles methods to Judge and Assessor classes to return a list of Award types.
* Adds Reports::Admin::AssessorJudgeAdminDataReport class to map csv headers and data.
* Adds Admin::UsersReportsController and route to handle download and adds link to index view.

* https://app.asana.com/0/1200504523179343/1206833235918007

* None

## ✅ Checklist

## 🖼️ Screenshots (if appropriate - no PII/Prod data):

Screenshot 2024-03-20 at 16 27 32

### DIFF
--- a/app/models/assessor.rb
+++ b/app/models/assessor.rb
@@ -67,7 +67,7 @@ class Assessor < ApplicationRecord
 
   def inactive_message
     !suspended? ? super : :suspended
- end
+  end
 
   def self.roles
     [["Not Assigned", nil], ["Lead Assessor", "lead"], ["Assessor", "regular"]]
@@ -146,6 +146,12 @@ class Assessor < ApplicationRecord
   def lead_roles
     FormAnswer::POSSIBLE_AWARDS.select do |cat|
       get_role(cat) == "lead"
+    end
+  end
+
+  def roles
+    FormAnswer::POSSIBLE_AWARDS.select do |cat|
+      get_role(cat) == ("lead" || "regular")
     end
   end
 
--- a/app/models/reports/user.rb
+++ b/app/models/reports/user.rb
@@ -1,0 +1,32 @@
+class Reports::User
+
+  def initialize(user)
+    @user = user
+  end
+
+  def call_method(methodname)
+    return "not implemented" if methodname.blank?
+
+    if respond_to?(methodname, true)
+      send(methodname)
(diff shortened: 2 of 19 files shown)


app/models/assessor.rb
```diff
@@ -149,6 +149,12 @@ def lead_roles
     end
   end
 
+  def roles
+    FormAnswer::POSSIBLE_AWARDS.select do |cat|
+      get_role(cat) == "lead" || "regular"
```
dreamfall, line 154: this is not gonna work, and always gonna return all roles.

app/models/reports/user.rb
```diff
@@ -0,0 +1,40 @@
...
+  def user_type
+    if @user.is_a?(Admin)
+      "Admin"
+    elsif @user.is_a?(Assessor)
+      "Assessor"
+    elsif @user.is_a?(Judge)
+      "Judge"
+    else
+      "User"
+    end
```
dreamfall, line 30: can do just `@user.class.name`

app/models/reports/user.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def awards_assigned
+    if @user.is_a?(Assessor) || @user.is_a?(Judge)
+      @user.roles.reject{|e| e=~ /promotion/}.join(", ")
```
dreamfall, line 35:
```suggestion
      @user.roles.reject{ |e| e=~ /promotion/ }.join(", ")
```

dreamfall (changes requested):

app/models/assessor.rb
TheDancingClown: Yes, I want to return the award types for which the assessor is either a regular or lead assessor. I want to omit awards they have not been assigned to.

dreamfall: @TheDancingClown i mean this condition is always true, does not matter what the role is or if there is one.

TheDancingClown: Ah I see - good spot, thanks

dreamfall (approved):

PR #2770, opened 2024-04-04
PO matrix questions improvements

Lubosky:

Modified matrix type of question. Required rows now take options (for now it will show in HTML form as data attribute) so we can have dependent rows which can be hidden if checkbox is not checked. Optimize JS a little as with the amount of input fields it was taking 2 seconds for 1 checkbox to change state. 

Asana card here: https://app.asana.com/0/1200504523179345/1205364676770329/f

None

## ✅ Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works


app/views/qae_form/_matrix_question.html.slim
```diff
@@ -26,10 +26,14 @@ table.matrix-question-table.govuk-table class="#{'auto-totals-column' if questio
             div.govuk-form-group
               label.visuallyhidden-on-desktop.govuk-label for=question.input_name(suffix: "#{x_heading.key}_#{y_heading.key}")
                 = "#{y_heading.label} numbers for #{x_heading.label}"
+
+              - required_row_data_opts = question.required_row_options
```
TheDancingClown, line 30: Is this variable used?

Lubosky: no, looks like I had some plans for it but didn't use it in the end. Thank you for noticing this.

PR #2855, opened 2024-05-02
Update README.md

matthewford:


README.md
```diff
@@ -1,73 +1,86 @@
...
+If you need to test collaborators editing the application at the same time, install [poxa](https://github.com/bitzesty/poxa).
 
-```
-ActiveRecord::StatementInvalid: PG::UndefinedFile: ERROR:  could not open extension control file "/usr/share/postgresql/9.3/extension/hstore.control": No such file or directory
-: CREATE EXTENSION IF NOT EXISTS "hstore"
-```
+### Installing Malware Scanning
 
-This means, that `hstore postgresql` extension needs to be installed:
+Files are uploaded to S3 and then scanned with ClamAV via the vigilion service.
```
Lubosky, line 55: Not sure what environment is this for as it's true only for production. Maybe mentioning that would make it clearer but it doesn't prevent user from setting things up.

Lubosky, line 55:
```suggestion
Files are uploaded to S3 and then scanned with ClamAV via the Vigilion service.
```

Lubosky (approved):
Just one typo and maybe one suggestion. Otherwise ok as I was able to follow the instructions to get it running from scratch.

PR #2968, opened 2024-06-18
Error fixes

DaniBitZesty:

* Corrects how 'step-errors' class is removed from side panel links to form sections, it was previously excluding links two-levels below elements with 'steps-progress-bar' class, so section C.1, C.2, C.3 links weren't targeted and the error icon persisted even after errors had been corrected

* Fixes missing errors for incomplete subfields on the initial render of the form. Previously, validation errors for hidden subfields were not displayed when navigating via the side panel links or the error summary links. These errors only became visible after navigating to another form section and then returning, this was because of `.filter(:visible)`, which excluded hidden elements from the validation process. 

* https://app.asana.com/0/0/1207573828239723/f
* https://app.asana.com/0/1200504523179343/1207573828239725/f

* 

## ✅ Checklist

### DIFF
--- a/app/assets/javascripts/frontend/form-validation.js.coffee
+++ b/app/assets/javascripts/frontend/form-validation.js.coffee
@@ -7,7 +7,7 @@ window.FormValidation =
     @validates = true
     $(".govuk-form-group--error").removeClass("govuk-form-group--error")
     $(".govuk-error-message").empty()
-    $(".steps-progress-bar .js-step-link").removeClass("step-errors")
+    $(".steps-progress-bar a.js-step-link").parent().removeClass("step-errors")
 
   clearErrors: (container) ->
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/frontend/form-validation.js.coffee
```diff
@@ -190,7 +190,7 @@ window.FormValidation =
 
   addSubfieldError: (question, subquestion) ->
     questionRef = question.attr("data-question_ref")
-    input = $(subquestion).find('input,textarea,select').filter(':visible')
+    input = $(subquestion).find('input,textarea,select')
```
dreamfall, line 193: I'm actually not sure about this, cause this somehow might impact some weird conditional questions if we have a question where the subfield is hidden 🤔

PR #3048, opened 2024-08-22
fix: Move AsimFormatter under Formatters module

TheDancingClown:

* This commit fixes the error `Unable to load application: NameError: uninitialized constant Formatters::AsimFormatter`


Lubosky (changes requested):
I think all references of `AsimFormatter.new` should be updated to `Formatters::AsimFormatter.new` to avoid similar errors `uninitialized constant AsimFormatter (NameError)`

Lubosky (approved):

PR #3108, opened 2024-10-17
Modify validations behaviour

sammo1235:
See [ticket](https://app.asana.com/0/1200504523179345/1207272963176223/f) for changes need, they do a good job of explaining the issues this PR attempts to solve. There is also a list of form questions this is relevant to (basically they are date, financial info, personal details and previous award wins).

* If form starts off empty, then we don't add validation errors for all the inputs the user has not yet touched.
* however, if we go back and delete an input, then it should show the validation error for that input.
* If the user navigates between different steps of the form (using save or the navigation menu on the left) then all validation errors should show, as before. However, now the behaviour has changed to only remove validation errors for the inputs that the user has touched, not clearing them wholesale like before.

There are two main modes to understand when reviewing the code - `@validate_only_touched_inputs` will be false when navigating around the different form steps (because we want to validate everything) and will be true once they start to interact with the inputs on a form (since we only care about validating what they touch).

* https://app.asana.com/0/1200504523179345/1207272963176223/f

* Very little, this is just JS changes

## ✅ Checklist

- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have squashed any unnecessary or part-finished commits

### DIFF
--- a/app/assets/javascripts/frontend/form-validation.js.coffee
+++ b/app/assets/javascripts/frontend/form-validation.js.coffee
@@ -1,5 +1,6 @@
 window.FormValidation =
   validates: true
+  validate_only_touched_inputs: true
   # validates numbers, including floats and negatives
   numberRegex: /^-?\d*\,?\d*\,?\d*\,?\d*\,?\d*\.?\,?\d*$/
 
@@ -10,8 +11,8 @@ window.FormValidation =
     $(".steps-progress-bar a.js-step-link").parent().removeClass("step-errors")
 
   clearErrors: (container) ->
-    if container.closest(".question-financial").length > 0
-      container.closest("label").find(".govuk-error-message").empty()
+    if container.closest(".span-financial").length > 0
+      container.closest(".span-financial").find(".govuk-error-message").empty()
     else if container.closest('.question-matrix').length > 0
       container.closest("td").find(".govuk-error-message").empty()
     else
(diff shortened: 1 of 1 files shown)


matthewford:
I like the fact that this is a JS change only, which means there should be little knock-on effects on updating the form in later workflow steps.

app/assets/javascripts/frontend/form-validation.js.coffee
```diff
@@ -286,18 +293,24 @@ window.FormValidation =
     questionDay = question.find(".js-date-input-day").val()
     questionDate = "#{questionDay}/#{questionMonth}/#{questionYear}"
 
+    questionYearTouched = !!question.find('.js-date-input-year.dirty').length
+    questionMonthTouched = !!question.find(".js-date-input-month.dirty").length
+    questionDayTouched = !!question.find(".js-date-input-day.dirty").length
+    allTouched = questionYearTouched && questionMonthTouched && questionDayTouched
```
sammo1235, line 299: There are probably cleaner ways of doing this in coffee

app/assets/javascripts/frontend/form-validation.js.coffee
```diff
@@ -31,10 +32,11 @@ window.FormValidation =
 
   appendMessage: (container, message) ->
     el = container.find(".govuk-error-message").first()
-    if el.text.length > 0
-      el.append(message)
-    else
-      el.append(" #{message}")
+    if not el.text().includes(message)
```
sammo1235, line 35: Since we now don't clear the entire question block for validation error before revalidating, we have to be careful not to add duplicate errors to the same elements

app/assets/javascripts/frontend/form-validation.js.coffee
matthewford:
```suggestion
[yearTouched, monthTouched, dayTouched] = ['.js-date-input-year', '.js-date-input-month', '.js-date-input-day'].map (cls) ->
  !!question.find("#{cls}.dirty").length
allTouched = yearTouched and monthTouched and dayTouched
```
It's not really much cleaner, though; I prefer the simplicity.

TheDancingClown (approved):

PR #3178, opened 2025-02-24
Confirm Environment variables

phil-l-brockwell:
[Asana](https://app.asana.com/0/1207904458299124/1208054457333363)


sammo1235 (commented):
There's code removal here that doesn't appear related to the ticket - is this correct?

sammo1235 (approved):

app/models/user.rb
```diff
@@ -158,10 +158,6 @@ def timeout_in
     ENV.fetch("SESSION_TIMEOUT", 24).to_i.hours
   end
 
-  def check_email_on_bounces!
-    ::CheckAccountOnBouncesEmail.new(self).run!
-  end
-
   def bounces_email_reason
```
matthewford, line 161: This also needs removing think; admins can see notify

PR #3188, opened 2025-03-07
Use correct keys when admin views or edits a PO/mobility form answer

sammo1235:
From the looks of `Reports::DataPickers::FormDocumentPicker` line 295-301 it seems like this key changes depending on the year that the form was submitted. I've done some checks in prod console and they seem to check out.

Tests needed some modification so the methods would produce the expected value.

### DIFF
--- a/spec/decorators/form_answer_decorator_spec.rb
+++ b/spec/decorators/form_answer_decorator_spec.rb
@@ -33,7 +33,7 @@ DOCUMENT_FIELDS = {
   innovation_desc_short: "innovation_desc_short",
   development_desc_short: "development_desc_short",
   development_management_approach_briefly: "development_management_approach_briefly",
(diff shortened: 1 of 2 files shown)


spec/decorators/form_answer_decorator_spec.rb
```diff
@@ -192,7 +192,7 @@
     describe "##{field}" do
       it "returns the document field with key #{DOCUMENT_FIELDS[field]}" do
         document = { DOCUMENT_FIELDS[field] => "An expected value" }
-        form = build(:form_answer, :development, document: document)
+        form = build(:form_answer, (field == :mobility_desc_short) ? :mobility : :development, document: document)
```
phil-l-brockwell, line 195: Could we put this logic in the factory?

phil-l-brockwell (approved):

PR #3209, opened 2025-03-26
Feedback after internal QA

sammo1235:


app/helpers/region_helper.rb
```diff
@@ -101,10 +101,197 @@ module RegionHelper
...
+    "South Glamorgan",
+    "West Glamorgan",
+    "Wrexham",
+  ].freeze
+
   def counties
     COUNTY_REGION_MAPPINGS.keys.sort
   end
 
+  def palace_invite_counties
```
phil-l-brockwell, line 291: Do we want to have different sets of counties for palace invites and users (or wherever else these are used) we should probably use one or the other in all places for consistency

app/models/palace_invite.rb
```diff
@@ -1,4 +1,5 @@
 class PalaceInvite < ApplicationRecord
+  POSSIBLE_TITLES = ["Mr.", "Mrs.", "Miss", "Ms", "Doctor", "Professor", "Sir", "Dame", "The Lord", "The Lady", "Lady"].freeze
```
phil-l-brockwell, line 2: Same here I think

app/helpers/region_helper.rb
sunnymander: @phil-l-brockwell it is intentional that the counties used here are slightly different. We're required to match them to a list of counties provided by the Palace for this form, as KAE will take the attendee details from this form and send them to the Palace.

phil-l-brockwell (approved):

PR #3222, opened 2025-04-01
Updates applicant dashboard banner

TheDancingClown:
- Removes the banner informing users of the email address change.
- Removes the settings dependent inset messages informing users of award season opening and closing.
- Adds banner notification informing users the awards are closed.


app/views/content_only/dashboard.html.slim
```diff
@@ -2,10 +2,10 @@
 
 - provide(:page_wrapper_class, "page-dashboard")
 
-= render "content_only/info_messages/email_phone_change_notification"
-
-- if Settings.after_current_submission_deadline?
+- if !Settings.after_shortlisting_stage?
```
sammo1235, line 5: Is this right? At the least it's a bit hard to read because of the `if not; else`. Is the award season closed if the we're not past the shortlisting stage?

TheDancingClown: Indeed - that is not right.

sammo1235 (approved):

PR #3235, opened 2025-04-04
Take bucket creds from vcap env

sammo1235:


lib/credentials_resolver.rb
```diff
@@ -24,4 +24,29 @@ def parse_database_config(config_json)
...
+  end
+
+  def clean_bucket_secret_access_key
+    s3_bucket_credentials(clean_bucket: true)[:aws_secret_access_key]
+  end
+
+  private
+
+  def s3_bucket_credentials(clean_bucket: false)
+    bucket_name = clean_bucket ? ENV["AWS_S3_PERMANENT_BUCKET"] : ENV["AWS_S3_TMP_BUCKET"]
```
phil-l-brockwell, line 47: You could just pass in `bucket_name` instead of having a switch here

phil-l-brockwell (approved):

